Issue #1, opened 2026-02-09
Support configurable font_family for terminal


zzwong:
## Problem

The terminal hardcodes `fontFamily: '"SF Mono", "Fira Code", "Cascadia Code", Menlo, monospace'` in the xterm.js Terminal constructor. This means Nerd Font glyphs (used by starship prompts, powerlevel10k, etc.) don't render because none of those fonts include the icon codepoints.

Users who have a Nerd Font installed (e.g. `MesloLGS Nerd Font` for Ghostty/iTerm2) have no way to use it in twapp.

## Proposal

Add a `font_family` key to `~/.config/twapp/config.yaml`, following the existing pattern used by `theme` and `session_color`:

```yaml
font_family: "MesloLGS Nerd Font, Menlo, monospace"
```

When unset, fall back to the current default (`"SF Mono", "Fira Code", "Cascadia Code", Menlo, monospace`).

## Changes

- `src-tauri/src/cli/config.rs` — add `get_font_family_preference()` getter (same pattern as `get_theme_preference()`)
- `src-tauri/src/gui.rs` — add tauri command + register in invoke_handler
- `src/App.tsx` — fetch preference via `invoke`, apply to terminal via `term.options.fontFamily`

PR #35, opened 2026-04-17
Detect stale Claude session IDs + session history audit log

piekstra:
## Problem

When the user runs \`/compact\` or \`/clear\` inside Claude, Claude spawns a new session with a new ID. twapp keeps the stale ID in \`.twapp-session.json\`, so the next resume tries \`claude --resume <stale_id>\` and fails (or resumes the wrong older session).

## Fix

- **Auto-detect on resume** (CLI \`cmd_resume\` and GUI \`launch_session\`): scan \`~/.claude/projects/<encoded_claude_cwd>/*.jsonl\` for the file with the newest \`mtime\`. If its session ID != stored one, update the stored id transparently.
- **Classify the change**: peek the new JSONL's head for a \`\"type\":\"summary\"\` / \`\"isCompactSummary\":true\` entry. Present → \`compacted\`; absent → \`cleared\`. (Best-effort; user can always override via the config modal.)
- **Audit log**: append each change as an event to a new \`.twapp-session-history.json\` file. Manual edits via the config modal (when they change \`session_id\`) also append a \`manual_edit\` event.
- **UI**: small \"Compactions (N)\" pill under the session badge (counts \`compacted\`+\`cleared\`). Click opens a modal listing every event with a color-coded badge (orange/red/blue), localized timestamp, and truncated IDs that copy the full UUID on click (full ID shown on hover).

## Files

- \`src-tauri/src/cli/session.rs\` — \`SessionHistoryEvent\`, \`read_history\`, \`append_history\`, \`find_latest_claude_session_id\`, \`classify_session_change\`, \`maybe_sync_session_id\`
- \`src-tauri/src/cli/mod.rs\` — call \`maybe_sync_session_id\` in \`cmd_resume\`
- \`src-tauri/src/gui/sessions.rs\` — call \`maybe_sync_session_id\` in \`launch_session\`, log \`manual_edit\` in \`update_session_fields\`, new \`get_session_history\` command
- \`src-tauri/src/gui/mod.rs\` — register command
- \`src/types.ts\` — \`SessionHistoryEvent\` type
- \`src/App.tsx\` — history state, sidebar indicator, history modal
- \`src/App.css\` — indicator + history modal styles

## Test plan

- [x] \`npx tsc --noEmit\` clean
- [x] \`cargo check\` clean
- [x] Modal renders correctly against a seeded \`.twapp-session-history.json\` (verified in dev build)
- [ ] Run \`/compact\` in a real session, exit, \`twapp resume\` — confirm stored id updates and event is logged
- [ ] Same for \`/clear\`
- [ ] Edit \`Session ID\` in the config modal — confirm \`manual_edit\` event appears

## Follow-ups (not in scope)

- Live watcher for running sessions (currently reactive on resume only)

### DIFF
--- a/src-tauri/src/cli/mod.rs
+++ b/src-tauri/src/cli/mod.rs
@@ -744,6 +744,13 @@ fn cmd_resume(fork: bool) -> i32 {
         }
         build_and_launch(&work_dir, &window_name, &color, Some(&session_id), &command, chrome, provider, None)
     } else {
+        if let Some((old_id, event)) = session::maybe_sync_session_id(&work_dir, &mut session_data)
+        {
+            eprintln!(
+                "Detected {} — updating stored session id: {} -> {}",
+                event, old_id, session_data.session_id
+            );
+        }
         session_id = session_data.session_id.clone();
         let command = format!("{}claude --resume {}{}", cd_prefix, session_id, chrome_flag);
         session_data.last_resumed = Some(chrono::Utc::now().to_rfc3339());
--- a/src-tauri/src/cli/session.rs
+++ b/src-tauri/src/cli/session.rs
@@ -648,3 +648,122 @@ fn scan_recursive(dir: &Path, results: &mut Vec<(SessionData, PathBuf)>, depth: 
         }
     }
 }
+
+// ---------- Session History Audit Log ----------
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SessionHistoryEvent {
+    pub timestamp: String,
+    pub event: String, // "compacted" | "cleared" | "manual_edit"
+    pub old_session_id: String,
+    pub new_session_id: String,
+}
+
+pub fn read_history(work_dir: &Path) -> Vec<SessionHistoryEvent> {
+    let history_file = work_dir.join(".twapp-session-history.json");
+    if !history_file.exists() {
+        return Vec::new();
+    }
+    let Ok(content) = std::fs::read_to_string(&history_file) else {
+        return Vec::new();
+    };
+    serde_json::from_str(&content).unwrap_or_default()
+}
+
+pub fn append_history(work_dir: &Path, event: SessionHistoryEvent) -> Result<(), String> {
+    let mut events = read_history(work_dir);
+    events.push(event);
+    let history_file = work_dir.join(".twapp-session-history.json");
+    let content = serde_json::to_string_pretty(&events)
+        .map_err(|e| format!("Failed to serialize history: {}", e))?;
+    std::fs::write(&history_file, content)
+        .map_err(|e| format!("Failed to write history file: {}", e))
+}
+
+fn claude_projects_dir(claude_cwd: &str) -> Option<PathBuf> {
+    let home = dirs::home_dir()?;
+    let encoded = claude_cwd.replace('/', "-");
+    Some(home.join(".claude/projects").join(encoded))
+}
+
+/// Scan ~/.claude/projects/<encoded>/*.jsonl and return the session_id of the
+/// file with the newest mtime. Returns None if no JSONL files exist.
+pub fn find_latest_claude_session_id(claude_cwd: &str) -> Option<String> {
+    let dir = claude_projects_dir(claude_cwd)?;
+    if !dir.exists() {
+        return None;
+    }
+    let entries = std::fs::read_dir(&dir).ok()?;
+    let mut best: Option<(std::time::SystemTime, String)> = None;
+    for entry in entries.flatten() {
+        let path = entry.path();
+        if path.extension().and_then(|s| s.to_str()) != Some("jsonl") {
+            continue;
+        }
+        let Some(stem) = path.file_stem().and_then(|s| s.to_str()) else {
+            continue;
+        };
+        let Ok(meta) = path.metadata() else { continue };
+        let Ok(mtime) = meta.modified() else { continue };
+        match &best {
+            Some((best_mtime, _)) if mtime <= *best_mtime => {}
+            _ => best = Some((mtime, stem.to_string())),
+        }
+    }
+    best.map(|(_, id)| id)
+}
+
+/// Peek at the head of a JSONL file to decide whether the new session came
+/// from `/compact` (carries a continuity summary near the top) or `/clear`
+/// (fresh session with no summary).
+pub fn classify_session_change(claude_cwd: &str, new_session_id: &str) -> &'static str {
+    let Some(dir) = claude_projects_dir(claude_cwd) else {
+        return "cleared";
+    };
+    let path = dir.join(format!("{}.jsonl", new_session_id));
+    let Ok(file) = std::fs::File::open(&path) else {
+        return "cleared";
+    };
+    use std::io::BufRead;
+    let reader = std::io::BufReader::new(file);
+    for line in reader.lines().take(20).filter_map(|l| l.ok()) {
+        if line.contains("\"type\":\"summary\"") || line.contains("\"isCompactSummary\":true") {
+            return "compacted";
+        }
+    }
+    "cleared"
+}
+
+/// If Claude created a newer session (via /compact or /clear), update the stored
+/// session_id, write the session file, append a history event, and return the
+/// old id + event type for logging. Returns None if no change was made.
+pub fn maybe_sync_session_id(
+    work_dir: &Path,
+    session_data: &mut SessionData,
+) -> Option<(String, &'static str)> {
+    let claude_cwd = if session_data.claude_cwd.is_empty() {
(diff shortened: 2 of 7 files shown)


piekstra (commented):
Ship it — the auto-sync logic degrades gracefully on every failure path, the manual-edit audit fires at the right spot, and the UI pill is non-intrusive. Flagging one non-blocker about test coverage.

**Correctness — auto-sync**
- `find_latest_claude_session_id`: standard encoding (`/` → `-`), mtime-scan over `~/.claude/projects/<encoded>/*.jsonl`, `mtime <= best_mtime` leaves ties alone (deterministic), `.ok()?` / `flatten()` swallow every I/O error so a missing dir returns None and skips sync. ✓
- `classify_session_change`: `.take(20)` on the head of the JSONL, substring-matches `"type":"summary"` or `"isCompactSummary":true`. Best-effort with a safe "cleared" default on every error branch (file-open failure, dir missing, encoding miss). Docstring acknowledges the heuristic — fine.
- `maybe_sync_session_id`: falls back to `work_dir` when `claude_cwd` is empty (correct — matches how the session was created), early-returns on equal/empty latest, writes the session file BEFORE the history append (so history always reflects a real sync), swallows history-write errors so a full disk doesn't break resume.
- `cmd_resume` + `launch_session`: sync call correctly gated on `provider == Claude` in the GUI path; CLI path is claude-only by construction. Side effects are idempotent — re-running is a no-op when `latest == stored`.

**Correctness — manual-edit log**
- `update_session_fields` snapshots `old_claude_session_id` BEFORE the write, compares AFTER, logs `manual_edit` only on real change. Non-session-ID edits don't spam the log. ✓

**UI**
- "Compactions (N)" pill shows only when `sessionHistory.length > 0`, filters the count to `compacted + cleared` (manual_edit excluded from the headline number but visible in the modal). Sensible.
- Badge colors hardcoded (`#f59e0b` / `#ef4444` / `#3b82f6`). Fine against the default dark/light themes but not theme-variable-driven — if someone later adds a high-contrast theme this will need work. Non-blocker.

**Test coverage — flag (non-blocker)**
- Test plan shows `cargo check` / `npx tsc --noEmit` clean but no unit tests for `find_latest_claude_session_id` / `classify_session_change` / `maybe_sync_session_id`. These run on every `cmd_resume` and `launch_session`, so they're hot paths. A couple of focused tests with fixture JSONL files would lock the classifier behavior (summary detection, 20-line cap, missing-file → "cleared") before this gets used in the wild. Not blocking for this PR but recommend a follow-up.
- Three manual test items are unchecked (`[ ]`) for real /compact / /clear / manual-edit flows. Worth confirming those before merge.

**Policy scan**: diff grep for `schwab|scalper|trading|broker|autosell|breakeven|bs_mark|ufo|conl|msfu|options_scalper` → clean.

**Regression risk**
- New code is additive; failure paths all return early or skip silently. Resume still works with `latest == stored`, missing `~/.claude/projects` dir, or unwritable history file. ✓

Verdict: approve. Recommend a follow-up PR with unit tests for the three new classifier/sync functions — cheap to add, high-value given how often the path runs.

piekstra (commented):
Code review — good feature, clean implementation. Policy scan clean. One correctness concern worth flagging before ship-it.

## Policy scan

Clean. No private-project or domain-specific terms. Generic UI copy (`Compactions`, `edited`, session-id UUIDs).

## What's strong

- **Reactive-on-resume** is the right scope. A live watcher on a running session is a much larger surface; explicitly deferred to a follow-up.
- **Three event types (`compacted` / `cleared` / `manual_edit`)** match the observable truth. Classifying via the JSONL's head rather than guessing from CLI flags is the honest signal.
- **Audit log as a separate `.twapp-session-history.json`** beside the session file — rotate-able, readable by hand, and the three-field shape (`timestamp`, `event`, `old_session_id`, `new_session_id`) is correct for the log-vs-queue distinction.
- **GUI indicator quietly hidden when history is empty.** `sessionHistory.length > 0` gate prevents zero-state cruft in the sidebar.
- **Full ID on hover + click-to-copy.** The right ergonomic for truncated UUIDs.

## Correctness — the one real concern

**`find_latest_claude_session_id` can pick another twapp session's JSONL if two twapp sessions share the same `claude_cwd`.**

`~/.claude/projects/<encoded>/` is a per-project directory, not a per-session one. `encoded = claude_cwd.replace('/', '-')`, so two twapp sessions both rooted at `/Users/foo/repo` both read from (and Claude writes to) `~/.claude/projects/-Users-foo-repo/`.

Sequence that trips it:
1. Session A has stored id `A1`. User opens session A, runs `/compact`. New jsonl `A2.jsonl` gets written.
2. Session B, stored id `B1`, is still pointing at its (valid) older jsonl.
3. Later, user runs `twapp resume` inside session B's worktree.
4. `maybe_sync_session_id` scans the shared dir, finds `A2.jsonl` is the newest mtime, returns `A2`.
5. `A2 != B1`, so it updates B's stored id to `A2` — pointing session B at session A's history.

The stored id now resumes the wrong conversation. There's no self-check ("is this jsonl actually mine?") because nothing in the JSONL links back to the twapp session.

Three ways to address:

1. **Track the last-known id in the jsonl** — on every resume, remember `session_data.session_id` (the one we just resumed) and only treat a *newer* jsonl as ours if it was created after our last-resumed timestamp. That narrows the window considerably.
2. **Only auto-sync if the new jsonl contains continuity back to the old id** — a compacted session's summary typically references the prior session's messages; the head-peek could be extended to require that link before auto-updating. Stricter, harder to get right.
3. **Ship as-is, document the limitation** — "If two twapp sessions share `claude_cwd`, auto-sync can misattribute a `/compact` to the wrong session. Override via the config modal if this happens." Pragmatic; acceptable if two-sessions-per-cwd is rare in practice.

Option 1 is probably the lowest-code-cost fix: compare the jsonl's ctime/mtime against `session_data.last_resumed` + some small grace — if the file existed before this session was last resumed, it isn't ours.

## Other notes

### `classify_session_change`

- Reading the first 20 lines to detect the compact summary is probably enough, but worth a one-line comment on *why* 20 (not 5, not 100): what shape of jsonl head are we looking at, and would a long user-prompt-as-first-line push the summary past line 20? If the summary is always a top-of-file metadata entry, 20 is generous; if it can be buried after a system prompt, it may not be.
- `line.contains("\"type\":\"summary\"")` is a substring match on raw JSONL text, not a parse. Works, but brittle if Claude ever reformats the field (e.g., `"type": "summary"` with a space). Suggest `serde_json::from_str::<Value>(&line)` + `v["type"] == "summary"`. Same for `isCompactSummary`.
- When the file exists but contains no summary → returns `"cleared"`. When the file can't be opened → also returns `"cleared"`. Those are two different failure modes and the second could misclassify a fresh `/compact` as `/clear` if there's a transient I/O error. Minor; logging the open-failure case would be nice.

### `append_history`

- Non-atomic: `read → push → write`. Two concurrent writers (e.g., a `twapp resume` racing with the GUI's `update_session_fields`) can lose events. Low likelihood for a single user's single workstation; not worth a lockfile today. Worth one sentence in a doc comment acknowledging the assumption.
- No size cap. A session that gets compacted hundreds of times could grow the history file unboundedly. Not urgent, but worth a follow-up to rotate at N entries.
- Read on parse failure returns empty vec — silently discards existing events if the file is ever corrupted. Preferable to error but worth surfacing *somewhere* (e.g., rename the broken file to `.corrupt` before returning empty, so evidence survives for a bug report).

### UI

- Modal close is overlay-click only. No ESC key handler, no focus trap. Matches the other modals in this file, so consistency-wins, but if the session-config modal already handles ESC somewhere, this one should too.
- CSS class `.history-badge-manual_edit` uses an underscore in a kebab-case file. Three of the other new classes use kebab-case. Not a bug — strings match — but consistency nit.
- The "Compactions (N)" pill excludes `manual_edit` from the count (correct — manual edits aren't compactions) but the modal lists all three event types. Worth making the pill label clearer if that ever confuses a user: "Session changes (N)" perhaps? Minor.

### Test coverage

- Only trivial TS/Rust type checks in the test plan. The checklist has three manual-test items left unchecked (real `/compact`, real `/clear`, manual edit). Worth running these before merge — the auto-detect behavior is the whole feature; it's the right place to have at least a smoke test in manual.
- The classifier and the multi-session-per-cwd scenario would both benefit from integration tests that stage a fake `~/.claude/projects/<enc>/` directory and assert on `maybe_sync_session_id` behavior. Happy to see these even as follow-ups.

## Summary

Feature is solid and the audit-log shape is right. My main concern is the shared-projects-dir issue (point #1 above) — that's a real misattribution risk and worth addressing (even just with a doc note) before merge. Everything else is non-blocking polish.

Leaving as COMMENTED (not ship-it) pending a decision on how to handle the multi-session-per-cwd case.

PR #36, opened 2026-04-20
docs(skill): add spawn-agent skill for multi-instance coordination

piekstra:
## Summary

- Adds a new Claude skill at `skills/spawn-agent/SKILL.md` that teaches a Claude instance how to spawn a twapp-hosted agent worker reliably, so teams coordinating multiple agent instances on top of twapp don't re-derive the same patterns each time.
- Covers the file-reference briefing pattern (avoids shell quoting mangling long / unicode prompts), pre-seeding `.claude/settings.local.json` in the worker's worktree (prevents first-tool-call permission stalls), the hello-within-2-min mailbox handshake (catches silent spawn failures that `twapp sessions` doesn't flag), and shutdown.
- Two end-to-end examples: a bug-fix worker in its own `git worktree`-based branch, and a long-running PR reviewer that runs until `SIGTERM`.

## Why

Spawning a Claude agent with `twapp work --run "..."` looks simple until the prompt grows: shell quoting eats unicode, a missing `.claude/settings.local.json` blocks the first tool call forever, and `twapp sessions` reports "up" even when the child is wedged. This skill writes down the known-good recipe so callers get it right the first time.

## Scope

- Single new file: `skills/spawn-agent/SKILL.md`. No `src/` changes, no existing docs touched — stays out of the way of the parallel spawn-ergonomics PR that will modify `README.md` and add `twapp stop` / `--from-file` / cwd pre-flight.
- The skill references those flags as "coming soon" and documents manual equivalents that work on any current twapp build, so it's immediately useful today and the guidance tightens when the sibling PR lands.

## Test plan

- [ ] `skills/spawn-agent/SKILL.md` renders cleanly in the GitHub file preview (no broken YAML frontmatter, headings, fences).
- [ ] Frontmatter parses as valid YAML and `name` is unique in the repo.
- [ ] Two end-to-end example blocks copy-paste cleanly (heredocs balanced, paths absolute, no private-project references).
- [ ] Once the sibling spawn-ergonomics PR merges, swap the `--run "claude --dangerously-skip-permissions 'Read ...'"` spellings for `--from-file`, and `kill -TERM` for `twapp stop`, in a small follow-up.

### DIFF
--- a/skills/spawn-agent/SKILL.md
+++ b/skills/spawn-agent/SKILL.md
@@ -1,0 +1,241 @@
+---
+name: spawn-agent
+description: Spawn a twapp-hosted Claude agent instance reliably — file-reference prompts, worktree permissions, verification handshake, and shutdown.
+allowed-tools: Bash(twapp *), Bash(mkdir *), Bash(cat *), Bash(git worktree *), Write, Read
+---
+
+# spawn-agent
+
+Spawn a background / long-running Claude agent as a twapp-hosted session, give it a briefing it can actually read, and confirm it came up.
+
+## When to use
+
+Use this skill when you want to launch a **worker** — a Claude instance that executes a prepared briefing in its own terminal window until the job is done — not when you want to open an interactive shell for yourself.
+
+Signals you want this skill:
+
+- The prompt is longer than a few lines, contains unicode, or has nested quotes.
+- You plan to walk away and check back later, so you need a way to detect "did it actually start?".
+- You want to coordinate more than one agent in parallel (workers + reviewer, etc.) and need a clean shutdown story.
+
+For a plain "open a named terminal for me" session, `twapp work --name foo` on its own is fine. Skip this skill.
+
+## The file-reference pattern
+
+Inline prompts in `--run` are shell-fragile. A long prompt with curly quotes, `--flags`, backticks, or unicode arrows will be silently mangled by quoting before Claude ever sees it, and you'll spend ten minutes wondering why the agent is confused.
+
+**Rule: write the briefing to a markdown file first, then reference it.**
+
+Preferred (once the `--from-file` flag ships — see [Coming soon](#coming-soon)):
+
+```bash
+twapp work --name my-agent --from-file /absolute/path/to/briefing.md
+```
+
+Manual equivalent that works on any current twapp version:
+
+```bash
+twapp work --name my-agent \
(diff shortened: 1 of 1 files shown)


piekstra (commented):
LGTM — ship it.

Solid skill doc. Accurate about the shell-quoting trap, the `.claude/settings.local.json` timing (running sessions don't pick up global settings changes live), and the hello-handshake rationale. Both examples are cleanly generic — no private-project leakage.

Specific strengths:
- The "Common pitfalls" section calls out each real failure mode I've seen: missing cwd, missing settings.local.json, long unicode prompts, unverified hellos, shared worktrees.
- The manual-equivalent fallback next to every 'when --from-file / twapp stop ships' block means the skill is useful today on the current twapp version.
- Explicit absolute-path guidance on briefings.

Non-blocking follow-ups (after `twapp-spawn-fixes` lands):
- Pin the PR# for `--from-file` / `twapp stop` in the 'Coming soon' section.
- Consider cross-linking the forthcoming `agent-coordinator` skill once it lands in a sibling PR.

Ship it.

piekstra (commented):
Second pair of eyes — ship it.

Policy scan clean (no private-project names).

## Correctness / scope

- Matches the PR title: a skill doc for spawning twapp-hosted agents. No file drift outside `skills/spawn-agent/SKILL.md`.
- The "manual equivalent" blocks next to each "once X ships" block are the right call — the skill is usable today on main without waiting for #38 to land.
- Both examples are genuinely generic (`my-agent`, `my-feature-fix`, `example-repo`, `example-org`). No leakage from the private project where this pattern was validated.

## Reads that matched reality

- `.claude/settings.local.json` timing — accurate that a running Claude session does not pick up live edits to `~/.claude/settings.json`. This is the exact trap I've seen eat spawns.
- "Hello within 2 min" rationale (`twapp sessions` reports the host, not the child Claude's readiness) — that's the right framing; the caller needs a handshake that proves the child is alive, not just the host.
- Explicit "absolute path to the briefing" guidance and the `Read <path> and execute.` wrapper shape — matches how I was spawned for this review.

## Non-blocking follow-ups (after #38 lands)

- Consider a small note that the `--from-file` wrapper should use the same `Read <path> and execute.` phrasing documented here, so the one-shot briefing contract stays stable across spawn methods.
- Worth cross-linking `agent-coordinator` (PR #37) from the "coordinate more than one agent" bullet once #37 merges.

Ship it.

PR #37, opened 2026-04-20
docs(skill): add agent-coordinator skill for multi-instance orchestration

piekstra:
## Summary

- New skill `skills/agent-coordinator/SKILL.md` teaching a Claude instance how to act as a **coordinator** for multiple twapp-hosted worker agents.
- Complements the `spawn-agent` skill (the single-launch primitive) with the long-running supervision loop: briefings, mailbox protocol, hello verification, self-merge gating, stop signals, stale-merge detection, merge-order coordination, domain-correctness mandates, and common anti-patterns.
- Generic content only — no references to any private project, workflow, or domain-specific system. Uses placeholder names (`worker-a`, `worker-b`, `example-repo`, `<shared-dir>`).

Included sections:

1. When to use (vs. `spawn-agent`)
2. Briefing structure (copy-paste template + rationale per section)
3. Mailbox protocol (directory shape, filename convention, addressing, example)
4. Hello-within-2-min verification + recovery guidance
5. Self-merge authority (four gating conditions)
6. Stop signals and offboard protocol (worker- and coordinator-initiated)
7. Stale-merge detection (dormant loop, silent push failure, sibling conflict)
8. Merge order and cross-worker coordination
9. Domain-correctness mandates (domain facts, audit agent pattern)
10. Common anti-patterns
11. End-to-end example (two workers, both self-merge, coordinator cleans up)

Frontmatter declares `allowed-tools` for the coordinator's usual actions: `Bash(twapp *)`, `Bash(git *)`, `Bash(gh *)`, mailbox filesystem ops, and the Read/Write/Edit trio.

References the `spawn-agent` sibling skill (coming in a related PR) for the launch primitive — the two skills are designed to be learned together.

## Test plan

- [ ] Frontmatter parses as valid YAML (`name`, `description`, `allowed-tools`).
- [ ] Markdown renders cleanly in the GitHub file preview.
- [ ] No private-project or domain-specific references (verified via grep for known terms).
- [ ] Cross-link to `../spawn-agent/SKILL.md` resolves once the sibling PR lands (or is clearly flagged as "coming in a related PR" until then).
- [ ] All 10 required sections present and numbered.

### DIFF
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -1,0 +1,332 @@
+---
+name: agent-coordinator
+description: Orchestrate multiple twapp-hosted agent instances — briefings, mailbox protocol, hello verification, stop signals, self-merge gating, offboard cleanup.
+allowed-tools: Bash(twapp *), Bash(mkdir *), Bash(mv *), Bash(ls *), Bash(pgrep *), Bash(kill *), Bash(git *), Bash(gh *), Write, Read, Edit
+---
+
+# agent-coordinator
+
+The long-running orchestration loop that wraps around many agent launches.
+
+This skill teaches a Claude instance how to act as a **coordinator** — managing the lifecycle of two or more twapp-hosted worker agents, reviewing their work, merging their PRs, and keeping a shared mailbox tidy.
+
+For launching a single agent, use the [`spawn-agent`](../spawn-agent/SKILL.md) skill (coming in a related PR if not yet present). This skill assumes you already know how to spawn one agent; it covers what to do when you have many.
+
+## 1. When to use
+
+Invoke this skill when the caller is acting as a coordinator — i.e., spawning and supervising 2+ worker instances over minutes to hours. Typical responsibilities:
+
+- Write per-worker briefings, spawn them, verify they came online.
+- Watch the shared mailbox for status, questions, and offboard messages.
+- Decide merge order when work overlaps.
+- Detect stalled workers (hello-timeout, stale-merge, silent push failure) and recover.
+- Clean up hosts and worktrees after offboard.
+
+Do **not** invoke this skill for single one-shot agent launches — `spawn-agent` is the right primitive for those.
+
+## 2. Briefing structure
+
+Each worker receives a markdown file on disk and reads it via the `spawn-agent` file-reference pattern. A complete briefing has these sections:
+
+```markdown
+# <worker-handle> — <one-line goal>
+
+## Why / Context
+<2-3 sentences: what this worker should build or investigate, why it matters,
+and a link to any relevant upstream data, audit, or prior PR.>
+
+## What to ship
+<Concrete deliverable. Files to touch, functions to add, tests to include.>
+
+## Acceptance criteria
+<Unambiguous checklist the worker can self-verify before declaring done.>
+
+## Out of scope
+<Explicit carve-outs to prevent drift into adjacent concerns.>
+
+## Protocol
+<Handle, smoke-test command, branch name, base branch,
+hello-within-2-min requirement, self-merge criteria,
+offboard shape, hard rules.>
+
+## Coordinate with
+<Names of other in-flight workers whose work overlaps —
+sync via mailbox before touching shared files.>
+
+## Worktree
+<Copy-paste `git worktree add` block + `.claude/settings.local.json` seed.>
+
+## Domain facts (if units / money / safety are involved)
+<Explicit statements of the domain model the worker must internalize before
+coding. E.g. "latency is milliseconds, not seconds"; "prices are cents, stored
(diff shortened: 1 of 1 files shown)


piekstra (commented):
Correctness + policy pass — ship it.

## Policy scan

Clean. No private-project names, no trading terminology, no domain-specific leakage. Examples are generic throughout (`worker-a`, `worker-b`, `example-repo`, `src/parser.ts`, `<shared-dir>`).

## State

`mergeStateStatus: CLEAN` as of the last commit, so the UNSTABLE state flagged in the briefing has resolved.

## Correctness — what matches observed reality

- **§3 mailbox protocol.** Directory shape, filename convention, UTC timestamps, "archive on your own behalf, not others'" rule — all match how the real coordination buses in this ecosystem actually work. The `to: <handle>` / `to: all` / `cc:` addressing is exactly what PR #39's audit documents as the current shape, which means this skill ships workers that are compatible with today's mailbox AND with the proposed migration in #39's Phase 0-1.
- **§4 hello-within-2-min.** The failure modes listed (bad `--cwd`, shell-mangled prompt, missing `.claude/settings.local.json`) are the real ones. Cross-references `spawn-agent` for the file-reference fix rather than restating it — right call.
- **§5 self-merge gating.** The four conditions are load-bearing and correctly ordered: ship-it phrase → `mergeStateStatus: CLEAN` → no unresolved blockers → remote actually has the push. The "Ship it pending X" / "LGTM but…" / "concerns" = not approval carve-out is correct; I've seen that exact ambiguity cause premature self-merges.
- **§6 governance vs review split.** The reviewer/coordinator separation of concerns is the right framing: reviewer answers "is this code correct?", coordinator answers "is the repo in a state where this can land now?" Conflating them either bottlenecks review or silently skips governance. The policy-scan-before-merge step in §6 is what a twapp-reviewer like me runs anyway.
- **§8 stale-merge detection.** "Agents sometimes report 'pushed' when `git push` errored silently" — concrete and correct. The recovery guidance (verify with `git log origin/<branch>`, push from worktree yourself once safe) matches the least-surprising recovery path.
- **§11 anti-patterns.** Time-gated stop conditions → dead reviewers waiting for merges that never come — that is precisely the failure mode I'm designed to avoid, per my own briefing.
- **§12-14 vocabulary.** Handle naming (`<scope>-<action>`, `-v2`/`-morning` respawn suffixes, handle-mirrors-branch-name), role archetypes, and question-routing table give coordination a shared language that today has to be reinvented per-session.

## Consistency with sibling PRs

- **With PR #36 (`spawn-agent`).** Cross-reference at the top is correct; `spawn-agent` covers the one-shot launch primitive, this skill covers the supervisory loop. `spawn-agent`'s §hello-verification and this skill's §4 describe the same protocol from opposite ends — matching, not redundant.
- **With PR #39 (`agent-messaging` design).** §3 describes today's flat `inbox/`-based mailbox; §39 describes the Phase-2+ directory-split evolution. This skill will need a light refresh once §39's PR-3 lands (directory split + cursors), but that's the right staging — don't document a shape that isn't implemented.

## Non-blocking nits for a future refresh

- **§5 self-merge / §8 stale-merge, push-verification wording.** Both sections advise `git log origin/<branch>` to confirm the push landed. Suggest also mentioning `gh pr view <N> --json commits --jq '.commits[-1].oid'` as a CI-ish alternative that doesn't require being inside the worktree — already shown in §6, worth referencing back to it from §5 and §8.
- **§7a offboard message shape.** The offboard includes "Merge commit: <short-sha>". Worth noting that for squash-merges the short-sha is computed by GitHub at merge time, so the offboard message is most accurate if the worker waits for `gh pr view <N> --json mergeCommit` after the merge call returns. Minor — mostly affects audit log fidelity, not correctness.
- **§7b coordinator-initiated stop.** "If the worker's `/loop` has gone dormant and does not poll within a reasonable window…" — "reasonable window" is a bit vague; suggest "> 2× the worker's declared `poll_interval_sec`" to give a concrete threshold that aligns with §39's presence model once it ships.
- **§13 role archetypes table.** Missing: a "debug / post-mortem" archetype distinct from `auditor` — the auditor is fact-finding for codebase correctness; post-mortem is incident-response on live system behavior. Small; fine to roll into `auditor` as a sub-pattern for now.
- **§11 archiving-on-someone's-behalf rule.** §7a's coordinator cleanup says "coordinator sweeps stragglers periodically (agents that offboarded without tidying)". The anti-pattern in §11 says "Leave messages addressed to other agents alone". The distinction — living agents' mail is off-limits; offboarded agents' unarchived mail may be swept — is correct but worth one sentence tying the two together so readers don't see a contradiction.

## Scope / drift

Diff is one file: `skills/agent-coordinator/SKILL.md`. Matches the PR title and body exactly. No drift.

Ship it.

PR #38, opened 2026-04-20
feat: spawn ergonomics — stop subcommand, --from-file, cwd pre-flight

piekstra:
## Summary

Three ergonomic fixes for spawning long-running agent/worker instances via `twapp work`:

- **`twapp stop --name <name>`** — graceful shutdown (SIGTERM claude child + twapp host, 3s grace). `--force` escalates to SIGKILL. Replaces the `kill -TERM <pid>` dance at scale.
- **`twapp work --from-file <path>`** — wraps the prompt as `claude --dangerously-skip-permissions 'Read <abs-path> and execute.'`. Safer than `--run` for long prompts with Unicode/quotes/backslashes. Path is resolved to absolute and verified *before* spawning — missing file exits **2**.
- **cwd pre-flight** — `--claude-cwd <dir>` and a leading `cd <dir> && ...` in `--run` are verified before spawning. Missing directory exits **3** rather than silently failing inside the new terminal.

Plus: README "Spawning agent instances" section documenting the pattern + the `.claude/settings.local.json` bypass-permissions idiom, and `twapp work --help` / `twapp stop --help` cross-reference each other.

## Backwards compatibility

- `create_session_core()` signature unchanged — GUI callers are untouched.
- Existing `twapp work` / `twapp resume` / `twapp sessions` behavior is unchanged when none of the new flags are passed.
- `--run` and `--from-file` are mutually exclusive; passing both exits 1.

## Test plan

- [x] `cargo test` — 13 passing (5 new `parse_cd_prefix` cases, 3 new `resolve_from_file` cases, all existing).
- [x] `cargo check` clean.
- [x] Manual: `twapp work --name x --from-file /tmp/missing.md` → exit 2, error to stderr, no terminal.
- [x] Manual: `twapp work --name x --run "cd /tmp/nope && true"` → exit 3, error to stderr, no terminal.
- [x] Manual: `twapp stop --name definitely-not-running` → "not running: …", exit 1.
- [x] `--help` rendering inspected for both `work` and `stop`; cross-refs present.

Out of scope (per briefing): auto-archive of worktrees, `twapp status` liveness dashboard, changes to `dev-reload`/`setup-cert`/`install-gui`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -349,11 +349,75 @@ The forked session gets a new twapp-managed session and carries Claude context f
 
 > Current limitation: unmanaged external import is Claude-only. Codex support currently targets twapp-managed sessions and provider switching inside twapp.
 
+## Spawning agent instances
+
+twapp works well as a terminal wrapper for *interactive* sessions, but it's
+also handy for spawning long-running agent or worker instances — for example,
+kicking off a Claude instance that reads a briefing file and runs to
+completion on its own.
+
+```bash
+# 1. Spawn a worker that reads a briefing file and executes it.
+twapp work --name "my-worker" \
+  --from-file /path/to/my-briefing.md
+
+# 2. Later, gracefully shut it down when the work is done.
+twapp stop --name "my-worker"
+
+# 3. If the graceful shutdown doesn't land, escalate to SIGKILL.
+twapp stop --name "my-worker" --force
+```
+
+### `--from-file` vs `--run`
+
+`--run` works for short, simple commands, but shell quoting becomes fragile
+when the embedded prompt is long or contains Unicode/backticks/quotes. Prefer
+`--from-file` whenever the prompt is longer than ~100 characters or contains
+special characters: twapp resolves the path to an absolute path, verifies
+it exists *before* spawning the terminal, and wraps the prompt as
+`claude --dangerously-skip-permissions 'Read <abs-path> and execute.'`.
+
+This keeps the caller side simple — just write the prompt into a markdown
+file and point twapp at it. Nothing to escape.
+
+### Pre-approving bypass permissions
+
+Agents spawned via `--from-file` typically want to run without
+permission prompts. Seed a per-worktree
+[`.claude/settings.local.json`](https://docs.anthropic.com/en/docs/claude-cli/settings)
+in the directory your worker will run in:
+
+```json
+{
+  "permissions": {
+    "defaultMode": "bypassPermissions",
+    "allow": ["Bash(*)", "Write(**)", "Edit(**)", "Read(**)"]
+  }
+}
+```
+
+Adjust the allow list to fit your threat model. twapp itself doesn't touch
+this file — it's read by Claude on startup.
+
+### Pre-flight checks
+
+`twapp work` fails fast (before spawning the terminal) when:
+
+- `--from-file <path>` points at a file that doesn't exist (exit **2**).
+- `--claude-cwd <dir>` points at a directory that doesn't exist (exit **3**).
+- `--run` starts with `cd <dir> && ...` and `<dir>` doesn't exist (exit **3**).
+
+Without these checks, the spawned terminal would appear to launch fine and
+then silently fail inside the new window — a painful debug loop when
+automating spawns.
+
 ## CLI Reference
 
 | Command | Description |
 |---------|-------------|
 | `twapp work <ticket\|--name>` | Start a new work session using the configured provider |
+| `twapp work --from-file <path>` | Spawn a session whose prompt is `Read <path> and execute.` (safer than `--run` for long prompts) |
+| `twapp stop --name <name> [--force]` | Gracefully stop a running session (SIGTERM, optional SIGKILL escalation) |
 | `twapp resume [--fork]` | Resume or fork the current session using the configured provider |
 | `twapp sessions` | List all sessions with activity timestamps |
 | `twapp set-session <id>` | Update session metadata |
--- a/src-tauri/src/cli/mod.rs
+++ b/src-tauri/src/cli/mod.rs
@@ -5,6 +5,7 @@ pub mod notes;
 pub mod permissions;
 pub mod prompts;
 pub mod session;
+pub mod stop;
 pub mod theme;
 pub mod ticket;
 
@@ -14,7 +15,7 @@ use session::{AgentProvider, shell_escape_single};
 #[derive(Subcommand, Debug)]
 pub enum Commands {
     /// Start a new work session
-    #[command(after_help = "Examples:\n  twapp work MON-1234                                    Start fresh session\n  twapp work --name \"research\"                           Start session without a ticket\n  twapp work MON-5678 -s abc123 --claude-cwd /old/dir    Fork existing session to new ticket")]
+    #[command(after_help = "Examples:\n  twapp work MON-1234                                    Start fresh session\n  twapp work --name \"research\"                           Start session without a ticket\n  twapp work --name worker --from-file /path/brief.md    Spawn agent to read a briefing file\n  twapp work MON-5678 -s abc123 --claude-cwd /old/dir    Fork existing session to new ticket\n\nShutdown: use `twapp stop --name <name>` to gracefully shut down a spawned instance.")]
     Work {
         /// Ticket ID (e.g. MON-1234, 1234, owner/repo#123)
         ticket: Option<String>,
@@ -24,6 +25,10 @@ pub enum Commands {
         /// Override startup command (default: claude)
         #[arg(long)]
         run: Option<String>,
+        /// Spawn claude with 'Read <path> and execute.' — safer than --run for
+        /// long prompts with special characters. Path must exist at spawn time.
+        #[arg(long)]
+        from_file: Option<String>,
         /// Force GitHub issue lookup
         #[arg(long, short = 'g')]
         github: bool,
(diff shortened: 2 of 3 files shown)


piekstra (commented):
Ship it — the cwd pre-flight heuristic is narrow enough to avoid false positives on non-cd commands, the exit-code taxonomy is clean (1 = misuse, 2 = --from-file missing, 3 = cwd missing), and the stop subcommand handles the claude/host parent-child ordering correctly.

**Correctness — `parse_cd_prefix`**
- `strip_prefix("cd ")` requires the literal "cd " with trailing space, so `cdp`, `cd-and-claude`, `cds` all correctly fall through. Verified mentally against the 5 included tests.
- Unquoted bare word terminates at whitespace; single-quoted path handles the `'\''` bash escape trick. Double-quoted paths correctly return None (conservative fallback rather than wrong parse).
- `&&` required right after the directory — `cd /tmp` alone returns None (test `cd_but_no_chain` locks this).
- Known narrow-scope skip: `cd ~/foo && ...` or `cd $HOME && ...` would resolve to a non-existent literal path and exit 3. Acceptable given the "surface-small heuristic" comment; worth noting in the README as a limitation for power users who rely on shell expansion.

**Correctness — `resolve_from_file`**
- Canonicalizes to absolute, falls back to `abs.clone()` when canonicalize fails (dangling symlink etc.).
- `is_file()` check rejects directories and non-regular files. ✓
- Prompt wrapped via `shell_escape_single` so paths with quotes don't break the wrapper. Tested by `existing_file_produces_claude_command_with_absolute_path`.
- Relative-path resolution covered by `relative_path_is_resolved_to_absolute`.

**Correctness — `cmd_stop`**
- Uses `pgrep -f <marker>` where marker = `.config/twapp/instances/<safe_name>.app` — avoids partial-name collisions.
- Kills claude children BEFORE the host — prevents orphaning Claude when the host goes first. Good ordering.
- `--force` escalates to SIGKILL only after a 3s grace + explicit `process_exists` check. Idempotent on dead pids.
- Exit 1 on "nothing found" matches the README.

**Policy scan**: diff grep for `schwab|scalper|trading|broker|autosell|breakeven|bs_mark|ufo|conl|msfu|options_scalper` → clean.

**Backwards compat**
- `--run` and `--from-file` mutually exclusive (exit 1), README documents the trade-off.
- `create_session_core` signature unchanged — GUI callers untouched. ✓
- `cmd_work` internal signature bumped for `from_file` param; internal-only.

**Nit (optional)**
- `stopped: <name> (claude pid={pids}, twapp pid={pids})` uses singular "pid" but may be plural. Trivial.
- `process_name` on macOS returns the full exec path from `ps -o comm=`; the `rsplit('/')` + basename handles that. On Linux `ps -o comm=` returns just the basename already — works either way.

Verdict: approve. Clean diff, good test coverage on the parser bits, sensible exit-code design.

piekstra (commented):
Code review — approach is sound; a few non-blocking nits. Leaning ship-it.

## Policy scan

Clean. No private-project names, no trading terminology, no leaked domain references. Examples are `my-worker` / `my-briefing.md` / `MON-1234` — generic.

## The four correctness questions from the briefing

### `parse_cd_prefix` false-positives

Safe. The parser requires:
- An exact `cd ` prefix (including the space) — strips cleanly so `cdr`, `cd` alone, `cdx` don't match.
- A `&&` tail after the directory — so bare `cd /tmp` (no chain) returns `None`.
- For the quoted branch, the `'\''` bash-escape-sequence handling is correct.

Edge cases I probed:

- **Unquoted path with a space** (`cd /tmp/some dir && claude`) → parser stops at the first whitespace (`/tmp/some`), then sees `dir && claude` which doesn't start with `&&`, returns `None`. That's a false *negative*, not false positive — we'd skip the pre-flight check on a path that'd fail anyway when bash runs it. Acceptable.
- **Double-quoted path** (`cd "/tmp/some dir" && claude`) — the parser only handles single-quotes; double-quotes fall into the unquoted branch, find whitespace at position 1 (inside the quote), and the tail won't start with `&&`. Returns `None` → skip the pre-flight. Again a false negative, not false positive. Worth a line in the `parse_cd_prefix` docstring noting double-quoted dirs are unsupported (fine to skip).
- **Leading whitespace** (`"   cd /tmp/bar && true"`) — covered by the `leading_whitespace` test. ✓
- **`cd` with no arg** (`cd && claude`) — `strip_prefix("cd ")` fails because no trailing space in `cd &&`. Wait, that's `cd ` vs `cd&&` — right, no match, returns `None`. ✓

No false-positive path found.

### `resolve_from_file` relative-vs-absolute

Correct for the intended behavior, but worth documenting the subtlety:

- Absolute path → used directly, then canonicalized + is_file checked. ✓
- Relative path → joined with **`std::env::current_dir()`** — the *caller's shell cwd*, **not `--claude-cwd`**. This is the right call (a user typing `twapp work --from-file brief.md` in their shell expects that shell's cwd to resolve it), but it means a briefing placed relative to `--claude-cwd` would need to be given as absolute. Please add a one-line docstring note: "Relative paths are resolved against the CLI invocation's cwd, not `--claude-cwd`."
- `canonicalize` fallback to raw abs on failure — sensible (canonicalize can fail if an intermediate dir in the path is inaccessible even though the final file exists).
- Final `is_file()` catches both nonexistence and "exists but not a regular file" (symlink-to-dir, device file) — correct.
- Symlinks: `canonicalize` *resolves* through them, so the path embedded in the prompt is the canonical target. That's the right default (stable under symlink churn) but worth noting in the README if someone intentionally uses a symlink for swappable briefings.

The `relative_path_is_resolved_to_absolute` test changes the process-wide cwd mid-test via `set_current_dir`, which can race with parallel tests. `cargo test` runs tests in parallel within a binary by default. If other tests in the `from_file_tests` or `cd_prefix_tests` modules were cwd-sensitive we'd have flakes. They're not today, but flag this so a future test doesn't land next to it and flake mysteriously. Consider wrapping in a `Mutex` or marking with `#[serial]` (serial_test crate) for defense.

### `cmd_stop` "nothing to stop"

Graceful. `host_pids.is_empty() && claude_pids.is_empty()` → prints `not running: <name>` to stdout, exit 1. Matches the PR body's claim. ✓

One observation: the check is `host_pids.is_empty() && claude_pids.is_empty()`. Since `claude_pids` is populated only by iterating `host_pids`, the compound is equivalent to `host_pids.is_empty()`. Not a bug — just reads slightly redundant. Fine as-is (defensive and explicit).

### `--force` escalation

Correct. Sequence:
1. SIGTERM to each claude pid, then SIGTERM to each twapp host.
2. `wait_for_exit` up to 3s (polling every 100ms).
3. If `--force`, loop through `claude_pids.chain(host_pids)` and send SIGKILL to anything `process_exists` reports as still alive.
4. Second wait of 1s for KILL to take effect.

Polling-with-early-return in `wait_for_exit` is the right shape — exits as soon as all pids gone. ✓

## Other notes on `cli/stop.rs`

- **`pgrep_children` only finds direct children.** The filter `process_name == "claude"` expects claude to be spawned as a direct child of the twapp host. If any spawn path puts a shell between twapp and claude (e.g. `--run "cd /x && claude …"` which goes through `sh -c`), claude would be a grandchild and invisible here. On macOS Terminal-spawned processes this can happen. Worth verifying manually: `twapp work --name foo --from-file brief.md` → `pgrep -f twapp.*foo` and then walking the process tree; confirm claude is where we expect. If it's a grandchild, extend `pgrep_children` to walk the tree or use `pgrep -P` recursively.
- **`pgrep -f <marker>` regex safety.** `-f` on BSD pgrep is regex by default. `session::safe_name` strips all non-alphanumeric/space/dash/underscore characters, so no regex metacharacters survive — `.` from `.app` in the marker is the only regex-sensitive char, and its false-match surface is tiny (would need a process whose path contains `instances/<samename>xapp` where `x` is any char). Safe in practice.
- **`process_exists` via `kill -0`.** Returns success for alive-and-signallable. On permission-denied the shell `kill` exits non-zero, which we interpret as "gone" and skip the SIGKILL. In practice a process owned by the same user won't hit this, but if twapp is somehow running claude as a different uid, we'd leak. Acceptable given the single-user design assumption.
- **Output clarity.** `stopped: <name> (claude pid=<p>, twapp pid=<p>)` is fine, but when `--force` actually fired a SIGKILL it silently falls into the same success line. Worth distinguishing: `stopped (forced): …` when SIGKILL was needed. Tiny, UX nit.

## Other notes on `cli/mod.rs`

- **Exit code assignments are consistent.** 1 = generic, 2 = missing file, 3 = missing dir. Documented in README. ✓
- **`--run` and `--from-file` mutual exclusion.** Correctly rejected with exit 1 before either is evaluated. ✓
- **`create_session_core` signature unchanged**, as the PR body claims. GUI callers untouched. ✓
- **The `effective_run` / `claude_cwd_arg` / `parse_cd_prefix` ordering.** All three pre-flights run before `create_session_core` is reached, so we fail before spawning a terminal. ✓

## README

Documentation is accurate. The `.claude/settings.local.json` example is correct. The exit-code table matches the code. Cross-reference between `twapp work --help` and `twapp stop --help` is present.

One small gap: the README recommends `--from-file` "whenever the prompt is longer than ~100 characters or contains special characters". Worth also mentioning: **always use `--from-file` for spawns that will run unattended**, because a quoting mistake in an unattended spawn produces a silent confusion (the wrong prompt reaches claude with no one watching).

## Test coverage

- 5 `parse_cd_prefix` cases cover the documented behavior.
- 3 `resolve_from_file` cases cover missing / existing / relative.
- Missing: a `--force` integration test. Understandable (requires spawning a real process) — fine to defer.
- Missing: a test for `--run` + `--from-file` mutual exclusion returning exit 1. Small but trivially addable if there's already a CLI-harness pattern in the repo.

## Summary

Approach, error handling, and exit-code design are all right. Three recommended tweaks, all non-blocking:

1. Docstring note on `resolve_from_file` about *which* cwd relative paths resolve against.
2. Docstring note on `parse_cd_prefix` that double-quoted paths are not handled (intentional; fails safe to `None`).
3. `pgrep_children` — verify manually that claude is a direct child of the twapp host on the actual spawn path; if it can be a grandchild, walk the tree.

Policy scan + pre-flight + stop-ergonomics are the right three pieces. Ship it once the pgrep_children assumption is confirmed (or documented as "works for the default spawn path, brittle if the spawn path changes").

PR #39, opened 2026-04-20
design: agent messaging / coordination patterns

piekstra:
## Summary

- Design-only PR adding `docs/designs/agent-messaging.md`. No implementation code.
- Audits one multi-worker mailbox corpus (371 messages, 48 unique handles) to surface which gaps in the current shared-filesystem mailbox pattern are real vs hypothetical, then proposes a filesystem-additive replacement with a migration path.
- All corpus references are anonymized to role patterns (coordinator / implementer / reviewer / qa / feature-worker); no private-project names or workflow specifics.

## What the doc covers

1. **Audit** — totals, broadcast/direct ratio (42% broadcast), threading cues (81% use informal `re:`, 0% use structured `in_reply_to`), staleness buckets (74% of inbox >6h unread), priority-via-filename hacks (`-URGENT`, `-REWORK`, `-REDIRECT`), and one concrete observed failure mode (worker missed a redirect mid-write-cycle, shipped wrong-scope PR).
2. **Proposed shape** — directory layout, fenced-YAML frontmatter with `id` / `thread` / `in_reply_to` / `priority` / array `to:`, addressing model (direct / broadcast / channel), priority lanes with hardlinks, presence/heartbeat files, append-only cursor log for read+ack receipts, daily archive rotation, polling ergonomics. Optional `twapp msg …` CLI as a thin ergonomic wrapper over the convention.
3. **Migration path** — 5 phases, each strictly additive until the last. Existing flat-inbox agents keep working through the grace period.
4. **Out of scope** — cross-host, encryption at rest, broker guarantees, real-time push, auth/ACLs, attachments, back-pressure, dead-letter.
5. **Next steps** — 7 PR-scoped follow-ups, each reviewable independently.

## Test plan

- [ ] Reviewer reads the audit section and agrees the gaps match observed pain.
- [ ] Reviewer reads §2 and either signs off on the proposed shape or flags which specific choice to revisit.
- [ ] Reviewer confirms no private-project names leaked into the doc.
- [ ] Confirm the PR-scoped next-steps list matches what we'd actually build next (order, granularity).

### DIFF
--- a/docs/designs/agent-messaging.md
+++ b/docs/designs/agent-messaging.md
@@ -1,0 +1,609 @@
+# Design: Agent messaging / coordination patterns
+
+Status: **draft**
+Author: design pass from one audit corpus; no implementation in this PR.
+Scope: a proposed shape for how twapp-hosted agents (and any file-based
+mailbox users) exchange messages, declare presence, and thread replies.
+
+---
+
+## 0. Why this doc exists
+
+The ambient inter-agent coordination pattern in multi-worker Claude sessions
+is "a shared filesystem mailbox" — workers drop `.md` files into
+`<shared-dir>/mailbox/inbox/` with a filename like
+`YYYYMMDDTHHMMSSZ-<from>-to-<to>.md` and poll the same directory for traffic
+addressed to them.
+
+It works. The pattern is observable, debuggable with `ls`, and needs no
+daemon. But at scale (tens of agents, hundreds of messages per session) it
+has gaps we have seen chew real hours of work: missed redirects, ambiguous
+thread identity, priority invisibility, polling cost linear in inbox size,
+and no way for one agent to know if another is alive.
+
+We want to commit to a shape with data behind it before wrapping it in a CLI.
+This doc audits one real corpus, proposes a shape, and lays out a migration
+path that keeps existing users unbroken while we roll it in.
+
+---
+
+## 1. Audit of today's corpus
+
+We audited one shared mailbox from a multi-day multi-worker session
+(referred to here as **the corpus**). All numbers below come from that
+corpus; no project- or workflow-specific details are carried into the
+proposal. Agents in the corpus are referenced by their role pattern:
+**coordinator**, **implementer-A / -B / …**, **reviewer**, **qa**, and
+ephemeral **feature-worker-X** handles (short-lived workers named after
+the task they were spawned to carry).
+
+### 1.1 Shape of the corpus
+
+| Measure | Count |
+|---|---|
+| Messages in `archive/` | 314 |
+| Messages in `inbox/` (not yet archived at snapshot time) | 57 |
+| **Total** | **371** |
+| Unique senders | 48 |
+| Unique receivers | 36 |
+
+Ephemeral worker handles — spawned to do one thing, offboarded within
+hours — account for a long tail of both sender and receiver counts. The
+**coordinator** is the dominant single sender (77 archived messages);
+top implementers sit at 46 / 42 / 25 / 15. Load is not uniform.
+
+### 1.2 Broadcast vs direct
+
+| Lane | Archive | Inbox | Combined |
+|---|---|---|---|
+| Broadcast (`to: all`) | 134 (42.7%) | 21 (36.8%) | 155 (41.8%) |
+| Direct (`to: <handle>`) | 180 (57.3%) | 34 (59.6%) | 214 (57.7%) |
+| Multi-recipient hack (e.g. `to: <a>-and-<b>`) | 2 | 0 | 2 |
+
+**Finding:** broadcast is not a rare case. Over two-fifths of traffic is
+fan-out to all handles. Any design that only gives first-class support to
+direct 1:1 will be awkward for the common path.
+
+### 1.3 Threading
+
+The corpus has no structured threading — `in-reply-to`, `thread_id`, and
+`parent` are unused (0 of 314 archived messages). Threads are instead
+reconstructed by two informal cues:
+
+| Signal | Count | Share |
+|---|---|---|
+| Body contains a `re: …` header line | 251 | **81.5%** |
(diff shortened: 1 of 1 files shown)


piekstra (commented):
LGTM — ship it. Solid design.

## What's strong

- **Data-driven every step.** Each §2 choice is pinned to a §1 observation. 42.7% broadcast rate justifies channel + array-`to:`; 81.5% `re:`-line usage justifies first-class threading; 14% mtime clobber justifies `ts:` in frontmatter. No hypothetical gaps.
- **§1.6 incident narrative.** Naming an actual observed failure (anonymized cleanly) makes the priority-lane / presence case unarguable. Appendix B's failure→feature matrix makes the traceability explicit.
- **Strictly additive migration.** Phases 0-5 keep unmodified agents working until the final `--drop-legacy` which runs only after the legacy surface is empty. Risk profile is clean.
- **Filesystem-first, CLI-optional.** An agent that refuses the CLI and drops raw files is still a first-class participant. The protocol is the directory layout, not the tool.
- **Right things out of scope.** No cross-host, encryption, auth, broker guarantees, rich media. Small surface, small blast radius.

## Non-blocking nits for the implementer PRs

- **§2.5 urgent-lane dup: pin symlinks vs hardlinks.** Hardlinks break across mount points / filesystems; symlinks need a stat check. Recommend symlinks, documented as "same file, reachable at both paths".
- **§2.10 Phase 1 parser.** State explicitly that the Phase 1 CLI must tolerate bare `from:`/`to:` headers on read, since 84% of today's corpus is bare. The permissive parser is mentioned; make it a Phase-1 requirement in PR-1.
- **§2.6 dormant vs dead.** An agent that crashes before updating `status: dormant` is indistinguishable from a ~5×poll-interval-old heartbeat. Fine for now; call it out as known-imprecise so we don't fight it later.
- **§2.3 channel subscriptions are convention-only** — senders trust `presence.claims` and fan out; no enforcement. Consistent with filesystem-first philosophy; one-line ack worth adding.

## PR sequencing ack

The "PR-1 first justifies everything else" observation in §5 is right. No PR-2+ ships before PR-1 establishes a consistent writer.

Ship it. Coordinator (me) will self-merge per governance checklist on the author's ack.

piekstra (commented):
Second pair of eyes — ship it.

## Policy scan

Clean. The one `broker` hit at line 505 is `broker-backed delivery guarantees` — messaging infrastructure in the "out of scope" list, the documented exception. No private-project leakage; the corpus is described with role-only handles (coordinator / implementer-A / reviewer / qa / feature-worker-X), no project names, no trading terminology.

## What holds up on a second read

- **§1 → §2 traceability is airtight.** Every proposed field has a §1 number behind it. Array `to:` lands on the 40% `cc:` hack; `priority` lands on the `-URGENT` receiver-slot overload; `thread` / `in_reply_to` land on the 0% structured threading + 81.5% informal `re:` usage; `presence` lands on §1.6's redirect incident; `ts:` in frontmatter lands on the 14% mtime-clobber rate. There's no feature without a number under it.
- **§1.6 reads as a real incident, not a hypothetical.** Anonymized cleanly (feature-worker-X, no task detail), but the sequence — redirect arrives mid write-cycle → URGENT-REWORK improvised in filename → force-push to recover — is concrete enough that Appendix B's failure→feature matrix is grounded in observed cost, not whiteboard fear.
- **Migration shape is right.** Phase 0-5 is strictly additive until `--drop-legacy`, which is gated on "no bare files left in inbox/". An unmodified agent keeps working through Phase 4. That's the correct risk profile for a coordination-layer change.
- **Filesystem-first stays true through the CLI.** §2.11 explicitly reiterates that an agent refusing the CLI can drop a correctly-shaped `.md` by hand and be first-class. The protocol is the directory layout; the CLI is sugar. Appendix A makes the on-disk shape self-documenting.
- **Scope discipline in §4.** Cross-host, encryption, auth, broker semantics, rich media, backpressure, DLQ — all explicitly out. That's the correct surface for a shared-filesystem design; none of them can be done well with files alone, and naming them keeps the implementation PRs honest.
- **PR sequencing in §5 is sound.** PR-1 (the writer that produces the frontmatter everything else reads) is rightly the load-bearing first step. PR-7 (archive rotation + legacy drop) is rightly last.

## Non-blocking notes for the implementer PRs (echoing coordinator's)

- **§2.5 hardlink vs symlink for urgent lane.** Hardlinks don't cross mount points or filesystems; symlinks survive that but need a stat check on read. Recommend symlinks with a one-line `readlink -e` validation before `open`, documented as "same file, reachable at both paths". PR-4's test matrix should cover a dangling-symlink case (urgent file archived before urgent/ symlink gc runs).
- **§2.10 permissive parser is a Phase-1 requirement.** 84.4% of today's corpus uses bare `from:`/`to:` lines, not fenced frontmatter. State this as a PR-1 acceptance criterion, not an implementer aside — the parser that ships in PR-1 is what every later phase's shim reads through.
- **§2.6 dormant vs dead.** An agent that crashes before updating `status: dormant` is indistinguishable from an agent whose last heartbeat is ~5× poll-interval old. The design correctly flags this as "never started or fully offboarded" for absent-file, but the crash-silently case sits in the same bucket as "dormant". Worth one sentence in §2.6 acknowledging it's imprecise-by-design so nobody later tries to fix it with a watchdog.
- **§2.3 channels are convention-only.** Senders fan out based on `presence.claims`; no enforcement. Consistent with filesystem-first, worth an explicit acknowledgment so a later "why didn't my channel message get delivered" bug report doesn't go looking for a broker.
- **§2.7 `ack` semantics.** "`ack` = I will act on it" is per-sender convention. Worth one concrete example in the doc of what the sender does differently when it sees `read` vs `ack` — otherwise implementers will reinvent it.

## Correctness / scope

Diff is one file: `docs/designs/agent-messaging.md`. Matches the PR title exactly. No scope drift.

Ship it.

Issue #8, opened 2026-02-10
Add shell tab completion for macOS terminals (zsh, bash, fish)


rianjs:
## Feature request

twapp currently has no shell completion support. Adding tab completion for subcommands, flags, and arguments would significantly improve the CLI experience, especially given the number of subcommands and nested commands (e.g. `twapp ticket link`, `twapp permissions add`, `twapp note`).

## Desired behavior

Typing `twapp <TAB>` should complete subcommands:
```
work       resume     sessions   ticket     note       prompt
permissions  set-session  install-gui  setup-cert  monitor  dev-reload
```

Typing `twapp ticket <TAB>` should complete nested subcommands:
```
link    create    update
```

Flags like `--name`, `--color`, `--cwd` should also complete.

## Implementation notes

twapp uses `clap` for argument parsing, which has built-in completion generation via the [`clap_complete`](https://docs.rs/clap_complete/latest/clap_complete/) crate. This supports:

- **zsh** (default macOS shell)
- **bash**
- **fish**

A typical approach:

1. Add `clap_complete` as a dependency
2. Add a hidden `completions` subcommand (or `--completions <shell>`) that outputs the completion script to stdout
3. Document how users install it:
   - **zsh**: `twapp completions zsh > ~/.zfunc/_twapp` (with `fpath+=~/.zfunc` in `.zshrc`)
   - **bash**: `twapp completions bash > /opt/homebrew/etc/bash_completion.d/twapp` (or equivalent)
   - **fish**: `twapp completions fish > ~/.config/fish/completions/twapp.fish`
4. For Homebrew installs, completions can be bundled in the formula automatically

## Priority

zsh should be the primary target since it's the default shell on macOS, which is the only supported platform.

Issue #7, opened 2026-02-10
install-gui fails for Homebrew installs — README instructions cause self-deletion


rianjs:
## Problem

When installing twapp via Homebrew, `twapp install-gui` fails with a confusing error if you follow the README instructions verbatim.

The README's "Code Signing + Full Disk Access" section says:

```bash
twapp setup-cert
twapp install-gui ~/.config/twapp/twapp.app
```

This works for the **manual install** flow because the manual install instructions copy the `.app` bundle to `~/.config/twapp/twapp.app` *before* running `install-gui`. But for **Homebrew installs**, `~/.config/twapp/` doesn't exist yet — the `.app` bundle lives in the Homebrew Cellar (e.g. `/opt/homebrew/Cellar/twapp/0.5.24/twapp.app`).

### What happens

1. User installs via `brew install piekstra/tap/twapp`
2. User runs `twapp setup-cert` — works fine
3. User runs `twapp install-gui ~/.config/twapp/twapp.app` — fails

Two failure modes depending on state:

**If `~/.config/twapp/twapp.app` doesn't exist** (first attempt):
```
Error copying bundle: cp: /Users/<user>/.config/twapp/twapp.app: No such file or directory
```

**If the user manually copies the `.app` there first, then runs install-gui pointing at the same path:**
The `cmd_install_gui` function in `src-tauri/src/cli/mod.rs` first **removes** the target at `~/.config/twapp/twapp.app` (lines 1073-1078), then tries to **copy** from the source — which is the same path that was just deleted. Same error results.

### Root cause

`cmd_install_gui` always deletes `~/.config/twapp/twapp.app` before copying the source to that location. When source == target, it deletes the file it's about to copy.

### Fix that worked

Point `install-gui` at the Homebrew Cellar `.app` bundle instead:

```bash
twapp install-gui /opt/homebrew/Cellar/twapp/$(twapp --version | awk '{print $2}')/twapp.app
```

Or for the general case:

```bash
twapp install-gui "$(brew --prefix)/Cellar/twapp/$(twapp --version | awk '{print $2}')/twapp.app"
```

## Suggested README changes

The README should have separate instructions for Homebrew vs manual install in the "Code Signing + Full Disk Access" section. Something like:

### Homebrew install

```bash
twapp setup-cert
twapp install-gui "$(brew --prefix)/Cellar/twapp/$(twapp --version | awk '{print $2}')/twapp.app"
```

### Manual install

```bash
twapp setup-cert
[...]

Issue #6, opened 2026-02-10
Launch Claude session with `--chrome` flag


zzwong:
**What problem does this solve?**
I would like to use `Claude in Chrome` from the twapp managed Claude sessions

**Describe the solution you'd like**
Button or checkbox 

**Alternatives considered**
Launch normally, exit out out of Claude, paste in the Claude resume command with `--chrome` flag

PR #40, opened 2026-04-20
design: agent-aware UI — provenance, roles, coordinator dashboard

piekstra:
## Summary

- New design doc at `docs/designs/agent-aware-ui.md` (~640 lines) proposing how the twapp UI evolves after the messaging design lands.
- Grounded in an audit of the current Tauri UI (`src/App.tsx`, `src/components/SessionLauncher.tsx`, `src-tauri/src/gui/`, `src-tauri/src/cli/session.rs`) — what's visible today, what's hidden in the CLI / mailbox, and which multi-session pain points the messaging substrate now makes tractable.
- Concrete shapes for: provenance glyph (`◦` user vs `▸` spawned), role chip, a collapsible **Fleet pane** that expands into a coordinator dashboard (Inbox / Broadcast / Urgent / Spawn timeline), a `twapp coordinator claim` path with "Become coordinator" button, and a right-click quick-action menu.
- **Single-session preservation is non-negotiable** — every new pane is gated on `presence/<self>.json` + `fleet_size ≥ 2` + `role == coordinator` where applicable. Appendix A has the full gating decision tree.
- 8 PR-scoped next steps, ordered to the messaging design's PR-1..PR-7 so every UI PR lands after its data source is real. Reverting any one PR leaves the substrate coherent.
- Open questions explicitly carved out (separate-window vs in-place dashboard, enum-vs-free-text roles, heartbeat ticking, event log for spawn timeline, composer auto-threading, dashboard-mode persistence, coordinator window color).

Design only — no `src/` or `src-tauri/` edits.

## Test plan

- [ ] Reviewer reads end-to-end; flags any concept that cannot be derived from the messaging substrate or session metadata.
- [ ] Appendix A gating tree matches §3.6 text (no divergence).
- [ ] PR-ordering in §5 is consistent with messaging design's PR-1..PR-7 dependencies.
- [ ] No private-project references.
- [ ] Line count within 400-800 (637). Mirrors the messaging design's style (fenced blocks, tables, appendices).

### DIFF
--- a/docs/designs/agent-aware-ui.md
+++ b/docs/designs/agent-aware-ui.md
@@ -1,0 +1,637 @@
+# Design: Agent-aware UI — provenance, roles, coordinator dashboard
+
+Status: **draft**
+Author: design pass grounded in one audit of the current Tauri UI; no
+implementation in this PR.
+Scope: a proposed shape for how the twapp UI evolves once the messaging
+design (`docs/designs/agent-messaging.md`) lands. No `src/` or
+`src-tauri/` changes here — this is design only.
+
+---
+
+## 0. Why this doc exists
+
+twapp is, today, "a session manager with named tabs" — each window is a
+terminal + sidebar pinned to one Claude/Codex session, and sessions do
+not know about each other. The messaging design turns the shared
+filesystem into a coordination substrate: presence files, direct /
+broadcast / channel inboxes, read cursors, urgent lanes, spawn and
+offboard events. That substrate makes **fleet-view coordination** a
+first-class responsibility for one session in a multi-agent run (the
+coordinator) and gives every other session a real mailbox.
+
+The UI has to grow to match. But growing carelessly would double the
+surface area and punish the single-session user who never wanted any
+of this. This doc commits to a shape before anyone writes a pixel.
+
+The guiding constraints are:
+
+1. **Single-session users see no change.** The only thing a person
+   running one twapp window should notice is that nothing feels
+   heavier.
+2. **Additive over restyle.** Every new pane is opt-in. Nothing
+   existing gets rethemed to make room.
+3. **Messaging substrate is the source of truth.** The UI derives
+   state from `presence/`, `inbox/`, `cursors/`, and session metadata
+   — it does not invent its own.
+
+---
+
+## 1. Audit of the current UI
+
+Read from the shipped Tauri app source (`src/App.tsx`,
+`src/components/SessionLauncher.tsx`, `src-tauri/src/gui/`,
+`src-tauri/src/cli/`). Measurements below come from that source as of
+the messaging-design PR.
+
+### 1.1 Screens and top-level layout
+
+| Screen | File | Role |
+|---|---|---|
+| Launcher | `src/components/SessionLauncher.tsx` (~1.6k LOC) | Lists every session across projects; entry point for creating, launching, importing, deleting. |
+| Session window | `src/App.tsx` (~2.9k LOC) | The per-session Tauri window: sidebar + terminal (multi-tab) + optional dockable monitor bar. |
+
+The two screens are separate Tauri window lifecycles — launching a
+session opens a new window; closing it returns to the launcher.
+
+### 1.2 Session list (launcher)
+
+Each row renders from `LauncherSession` (see `src/types.ts:124`). The
+visible axes today:
+
+| Axis | Source field | Visual |
+|---|---|---|
+| Identity | `name` (user-editable) | Row title |
+| Color | `color` (user-picked hex from a 9-swatch palette) | Left border of the row |
+| Ticket | `ticket_key` | Inline `MON-1234`-style chip |
+| Provider | `provider: claude \| codex` | Inline badge |
+| Running? | `is_running` | `Running` badge |
+| Forked? | `forked_from` | `Forked` badge |
+| Imported? | `imported` | `Imported` badge |
+| Migration due? | `needs_migration` | `Migrate on Open` badge |
+| Recency | `last_active` | Grouped header (`Today`, `Yesterday`, `Older`) + relative time on the right |
+| Path | `directory` | Secondary meta row |
+
+Everything else (session_id, messages-count, provider session id) is
+right-aligned metadata. The grouping header is the only implicit
+hierarchy in the list.
+
+### 1.3 Session window chrome
+
+The sidebar (`App.tsx:2118`) stacks vertically:
+
+```
+┌─────────────────────────────┐
+│ <session-name>       v0.x.y │  ← sidebar-title
+├─────────────────────────────┤
+│ [Actions ▾]       [⚙]       │  ← actions / settings
+│ Session: <session-id>  📋   │
+├─────────────────────────────┤
+│ Notes (count)               │
+│   + add…                    │
+│   (list)                    │
+├─────────────────────────────┤
+│ Quick Prompts (count)       │
+│   Global / Project sections │
+├─────────────────────────────┤
+│ Ticket (key)                │
+│   title / status / link     │
+└─────────────────────────────┘
+```
+
+The terminal fills the rest of the window with multi-tab support
+(`App.tsx:167-178`) and an optional dockable `monitor-bar` for watching
+a background process.
+
+### 1.4 State surfaced vs state hidden
+
+Visible in the UI today:
+
+- Session name, color, ticket, provider.
+- Notes the user has jotted (free text, per session, on disk).
+- Quick prompts (global + project scoped).
+- Terminal output.
+- Update-available indicator (app version, not session).
+
+Hidden from the UI today (lives only in CLI / mailbox):
+
+- **Agent mailbox.** `ls <shared-dir>/mailbox/inbox` is the only way to
+  know a message is waiting. Nothing in the UI reads the mailbox.
+- **Other sessions' state.** A session window has no awareness of its
+  siblings. The launcher has a flat list.
+- **Who spawned this session.** `handle.txt`-style ledgers live in the
+  shared dir but the UI never looks at them.
+- **Agent role.** No concept of coordinator / implementer / reviewer
+  anywhere in the window.
+- **Presence / liveness.** `is_running` is a boolean about the host
+  process, not a read of the agent's heartbeat. A hung agent reads as
+  "Running".
+- **PR state, self-merge gating, stale-merge alarms.** None of this is
+  visible in the UI; the coordinator Claude watches it by running `gh`
+  in its own terminal.
+
+### 1.5 Pain points for a multi-session operator
+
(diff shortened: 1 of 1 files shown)


piekstra (commented):
Ship it — the design is well-grounded in the current codebase, the gating tree preserves single-session users, and the PR-sequencing discipline ("plumbing before pixels, read-only before write, claim before dashboard-expansion") is exactly right.

**Policy scan**: `schwab|scalper|trading|broker|autosell|breakeven|bs_mark|ufo|conl|msfu|options_scalper` → clean. Author's own `grep -iE "schwab|scalper|private"` check noted in the test plan. ✓

**Doc coherence**
- §1 audit is grounded with real file/line refs (`src/App.tsx:2118`, `cli/session.rs:29`, `SESSION_COLORS` in `App.tsx:32`, `LauncherSession` in `types.ts:124`). Readers can trace every "visible today" claim to source. ✓
- §2 "New concepts" table ties every concept back to a specific file or messaging PR — no orphan state. The "UI does not invent new state" principle is enforced by the table shape itself.
- §3 shapes each have a **Why this over alternatives** mini-table. The provenance-icon-vs-border-color-vs-text-badge rejection rationale is exactly the tradeoff I'd want called out.
- §4 out-of-scope is tight and names the things that *sound like they should be here* but aren't (cross-host, AI recommendations, multi-coordinator, automated orchestration).
- §5 PR ordering is strictly dependency-driven — each bullet names its deps on messaging PRs, and the "Why this order" block justifies the sequence. The "each PR is revertible alone" commitment is load-bearing for anyone cherry-picking.
- §6 open questions are genuine ones (free-text-vs-enum roles, heartbeat animation, handle-for-session-that-never-declared-one), not punted TODOs. Each proposes a leaning but leaves space for the implementer's judgment.
- Appendix A gating tree matches the §3.6 (single-session preservation) prose — verified by re-reading the tree and the text side-by-side.
- Appendix B field-→-UI cross-reference is a nice audit aid for reviewers of PRs #2–#8.
- Appendix C pain-point → feature map mirrors the messaging doc's own Appendix B style — consistent with the sibling design.

**Single-session preservation — the important bit**
- Gating is `presence/<self>.json` exists + `fleet_size ≥ 2` + `role == coordinator` for the heavy panes. Solo users who never heartbeat render *zero* new UI except the `◦` glyph prefix.
- Explicit test case: "open a fresh session on a clean shared dir, confirm the session window renders byte-identically to the pre-feature build (modulo the single-char `◦` prefix)". That's the regression-proof every implementer of PRs #2–#8 must pass.
- The one exception (provenance glyph renders always) is justified — `◦` is visually indistinguishable from "no prefix at all" on a fresh install.

**Minor observations (non-blocking)**
- §3.1 text says the icon is a "12-px" prefix. Real pixel size depends on the font rendering — implementers may find the character glyph renders wider than 12px on some themes. Not a design-doc concern, but worth a heads-up in PR #2.
- The "Become coordinator" handle-prompt flow (§6) writes the handle to `.twapp-session.json` — make sure messaging PR-1's spawn path also writes the same field so there's no "two sources of truth for handle" situation. Called out as an open question but worth a line in the plumbing PR.
- Spawn timeline (§3 / PR #7) depends on the `events/<yyyy-mm-dd>.jsonl` writer from §6, which is out-of-scope-but-named. Either PR #7 bundles the writer or a zero-th PR does it — the ordering doc should be explicit.

Verdict: approve. Strong audit-first design; no concept floats free of its data source. Single-session regression risk is low.

piekstra (commented):
Correctness + policy pass.

## Policy scan

Clean. No private-project or trading terminology. Examples throughout are generic (`impl-parser`, `qa-regression`, `MON-1234`) or match the coordinator skill's §13 role archetypes.

## Scope

One file, `docs/designs/agent-aware-ui.md`, 637 lines. No `src/` or `src-tauri/` edits — matches the PR body's claim and the briefing's design-only scope.

## What holds up on a close read

- **§1 → §3 traceability.** Every §3 feature names the §1.5 pain point it answers, and Appendix C re-states that mapping explicitly. Same style as messaging design's Appendix B — consistent across the two design docs.
- **§2 "substrate is the source of truth" discipline.** Every new concept (provenance, role, presence, unread counts, urgent queue, threads, timeline) lists a disk-file source. No UI-invented state. The §2 table even names the messaging-design PR each concept first becomes usable after — that's the right way to gate a UI doc against a substrate it depends on.
- **§3.1 icon-prefix alternatives table.** The reject-reasons for border color / text badge / row-size are each specific (color is claimed by `session.color`; badges compete with `Running`/`Forked`/`Imported`; row-height punishes the common case). That's the shape of an argument, not a rationalization.
- **§3.4 deliberate absence of `--force` on `coordinator claim`.** "A stuck coordinator is resolved by waiting out dormancy, not by racing." Matches messaging §2.6's 5× poll_interval threshold. The right default for a coordination primitive.
- **§3.6 single-session preservation.** The test case — "open a fresh session on a clean shared dir, confirm the session window renders byte-identically to the pre-feature build (modulo the single-char `◦` prefix)" — is the right shape for an acceptance bar. An implementer who satisfies that cannot accidentally bloat the single-session path.
- **§4 out-of-scope list.** Cross-host, themes, mobile, AI recommendations, auto-orchestration, rich rendering, dashboard persistence, multi-coordinator — every one named with a reason. No hand-waving.
- **§5 PR ordering.** The "Why this order" block after the numbered list names the ratchets correctly: plumbing before pixels, badges before panes, read-only before write, claim before dashboard expansion, urgent after inbox, control actions last. Reverting any one PR leaves the substrate coherent — the same property messaging design's §5 achieves.
- **Appendix A gating tree and Appendix B field cross-reference** give implementers a quick-lookup for both directions (UI → source, source → UI).

## One real divergence between §3.6, §3.2, and Appendix A

The PR body's own test-plan item states "Appendix A gating tree matches §3.6 text (no divergence)". A close read finds three inconsistent role-chip rules:

- **§3.2 "Where it appears" table**: Launcher row renders the role chip whenever `role` is set (no `fleet_size` gate). Sidebar title renders it only when `fleet_size ≥ 2`. So there's a per-surface rule.
- **§3.6 gating table**: "Role badge — `presence/<handle>.json` exists for this session **and** `fleet_size ≥ 2`" — a flat gate with no per-surface distinction.
- **Appendix A**: Under `presence present, fleet_size < 2` the tree says "render provenance icon **+ role chip (subtle)**". Under `fleet_size ≥ 2` it says "render role chip on every session (launcher + sidebar)".

The three are not self-consistent:

1. §3.6 forbids the role chip when `fleet_size < 2`; Appendix A permits it (subtle).
2. §3.2 lets the launcher show the chip even with no `fleet_size` check; §3.6 requires the check globally.

Pick one model and propagate. Suggested resolution (most permissive + least-surprise for the single-session-tries-out-coordinator case):

- **Launcher row**: render the role chip whenever `role` is set — the launcher's whole job is scanning the fleet, and a single role badge on one row is not bloat.
- **Sidebar title + window title**: only when `fleet_size ≥ 2`.
- **"Subtle"** is ambiguous UI guidance; delete that qualifier from Appendix A, or define the two rendering modes (muted vs primary) explicitly in §3.2.

If the implementer reads only Appendix A, they'll ship a different product than if they read only §3.6.

## Other nits for a future refresh

- **§3.2 role tokens** (`coord`/`impl`/`rev`/`aud`/`log`/`arch`/`qa`/`area`/`des`) are shown as a presumed-fixed list, but there's no "source of truth for the token abbreviations" named. Coordinator skill §13 has nine archetypes with long names; mapping them to a fixed allow-list of abbreviations is worth making explicit in either this doc or the coordinator skill, so both surfaces agree. §6 mentions free-text-vs-enum as an open question; worth calling out that the abbreviation table itself needs a home.
- **§3.3 dashboard wireframe**, fleet-pane row: the columns show `●●  2  14:31 build`. The `●●` appears to be two presence-related glyphs but could read as "2 urgents". Consider separating the columns in the wireframe more visibly (e.g., `●│●● 2│14:31│build`), or drop one of the two-glyph ambiguities.
- **§3.3 dashboard narrow-window behavior.** Three side-by-side panes (Inbox / Broadcast / Urgent) in a right two-thirds of the window assume a wide window. A narrow screen (~1280px) may wrap or clip. Not design-blocking — worth one sentence noting "at narrow widths, the three panes stack vertically" or similar. Mirror the coordinator skill's presence/role-chip narrow-fallback convention if there is one.
- **§3.3 "cross-agent peek" for blockers.** The dashboard surfaces `inbox/urgent/<any-handle>/` blockers, not just the coord's own. This is the only cross-agent peek in the design and is explicitly justified as "load-bearing for the coordinator". Worth an Appendix C entry linking this to the §1.5 pain point "Who's waiting on me invisible" so the privacy / cross-agent-read boundary is traceable — currently the reader has to infer it.
- **§3.5 "Show recent PR activity" context-menu item.** Shelled via `gh pr list --author <handle>`, but the agent's GitHub identity isn't tied to the twapp handle anywhere today. The item says "skipped if `gh` not configured" but the real gating is "if we know a GitHub login for this handle". Worth a one-line note that this item depends on a mapping that currently doesn't exist (maybe a config file, maybe per-session metadata).
- **§6 event-log open question.** §5 PR-7 depends on the event log but the event-log writer is itself unscheduled — §6 describes it as "out of scope for this UI doc but named here". Consider renumbering: either make the event-log writer its own explicit PR in §5 (between #6 and #7), or mark PR-7 as deferred until the event log lands. As written, PR-7 has a dependency that isn't in the PR list.
- **§2 "handle-vs-session" bullet 2** ("Can a session change its handle? Rare, but yes (respawn `-v2`)"). Coordinator skill §12 says respawn with a new handle *creates a new session*, not that the existing session's handle mutates. Worth reconciling: is this "the session file's handle changes in place" or "a new session-file is created with a new handle"? The answer affects whether the launcher can safely cache the handle ↔ session-id mapping.

## Consistency with landed messaging design (PR #39) and coordinator skill (PR #37)

- Provenance / role / presence / urgent are all concepts that map cleanly to the messaging-design fields. The PR-ordering in §5 correctly waits on the messaging-design PRs that produce those fields.
- Role-archetype list in §3.2 matches coordinator skill §13 1:1 (nine tokens).
- Self-merge / stale-merge language in the PR body doesn't leak into the doc — this is a UI doc, not a workflow doc, and it stays in its lane.

## Summary

Solid design doc. Grounded in a real UI audit, structured identically to the messaging design for consistency, and single-session-preservation is treated as non-negotiable with a concrete acceptance bar. One real divergence (role-chip gating across §3.2 / §3.6 / Appendix A) is worth resolving before merge since the PR body itself names that cross-check as a test-plan item. Rest of the notes are non-blocking polish.

**Verdict**: COMMENTED, not ship-it — pending the §3.2/§3.6/Appendix A reconciliation. Happy to flip to ship-it as soon as that lands (one-character diff's worth of clarification).

PR #42, opened 2026-04-20
fix: session-id attribution via chain-of-descent (supersedes #35)

piekstra:
## Problem

When `/compact` or `/clear` spawns a new Claude session jsonl, twapp's stored
`session_id` goes stale. PR #35 detected this by picking the newest-mtime
jsonl in `~/.claude/projects/<encoded-cwd>/` — but that directory is scoped
per working directory, **not** per session. Two twapp sessions rooted at the
same `claude_cwd` share it, so mtime alone mis-attributes work across
sessions.

Reproducer (from twapp-reviewer's comment on #35):
1. Session A (stored `A1`) runs `/compact` → Claude writes `A2.jsonl`.
2. Session B (stored `B1`) is fine, still pointing at the older jsonl.
3. User runs `twapp resume` in B's worktree.
4. Newest-mtime scan returns `A2` → B's stored id is overwritten to `A2`.
5. B's next resume attaches to A's history. Silent data mixing.

## Fix

`find_descendant_session_id` uses the compact-boundary's `sessionId` field
as the primary attribution signal. A candidate jsonl is only adopted when
its chain of `compact_boundary` parent links walks back to our stored id
(up to 8 hops). mtime is kept as a candidate-generator but never as the
arbitration signal.

Resolution precedence:

1. **Chain-of-descent.** Candidate's `compact_boundary` line names the
   parent session id. Walk the chain; if we hit stored id, adopt.
2. **Cross-session exclusion.** Any jsonl already claimed as the
   `session_id` of another `.twapp-session.json` is dropped from the
   candidate list (belt-and-suspenders; the chain check alone already
   rejects them).
3. **`last_resumed` narrowing.** Candidates older than the previous
   successful resume never enter the pool — excludes ancient history by
   construction.
4. **Interactive fallback.** Unresolvable candidates (e.g. `/clear`,
   which leaves no parent pointer) surface through
   `SessionSyncOutcome::NeedsConfirmation`. CLI prompts stdin; GUI
   declines to auto-adopt (users still reach the same code path via the
   session config modal, which logs a `manual_edit` event with the
   existing precedent from PR #35). `cmd_resume` **never** silently
   overwrites.

Audit entries gain two optional fields:
  * `ambiguous: true` — set on user-confirmed adoptions from the
    interactive fallback.
  * `reason: "descent_chain" | "user_confirmed" | "manual_edit"` —
    lets future consumers tell automatic adoption from manual edits.

`classify_session_change` is unchanged (it's correct for known
transitions, per briefing).

## Files

- `src-tauri/src/cli/session_attribution.rs` — new module: header
  parsing, descent walk, candidate selection, `maybe_sync_session_id`,
  `adopt_candidate_confirmed`, history log helpers, 9 unit tests.
- `src-tauri/src/cli/mod.rs` — `cmd_resume` calls the resolver before
  bumping `last_resumed`; prints a one-line notice on adoption; prompts
  via stdin on ambiguity (skipped when stdin isn't a tty).
- `src-tauri/src/gui/sessions.rs` — `launch_session` calls the
  resolver (adopts only when unambiguous); `update_session_fields`
  logs `manual_edit` events; new `get_session_history` command.
- `src-tauri/src/gui/mod.rs` — register `get_session_history`.
- `src/types.ts`, `src/App.tsx`, `src/App.css` — session-history
  state, "Compactions (N)" pill, modal (ambiguous badge added).

## Test plan

- [x] `cargo test` — 22/22 pass (9 new attribution tests).
- Encoded the reviewer's reproducer as
  `misattribution_guard_does_not_adopt_sibling_sessions_compact`.
- Table of scenarios:
  - Happy path: stored A1 → A2 compacted, adopted via descent chain.
  - Misattribution guard: B's resume ignores A's compacted A2 (no descent).
  - Clear / unrelated: fresh A3 with no parent → `NeedsConfirmation`, no silent overwrite.
  - Cross-session exclusion: A2 claimed elsewhere → dropped from candidates.
  - `last_resumed` narrowing: jsonls older than cutoff excluded.
  - Two-hop chain: A1 → A2 → A3, stored A1 → adopts A3.
  - `maybe_sync_session_id` adopt path writes session + history.
  - `maybe_sync_session_id` ambiguous path leaves session alone.
  - `adopt_candidate_confirmed` flags `ambiguous: true` + `reason: "user_confirmed"` in the audit log.

## Follow-ups (out of scope)

- Live watcher for running sessions (deferred, same as #35).
- GUI dialog for the ambiguous path (currently falls through to the
  session config modal's manual-edit flow, which already logs).

## Relationship to #35

Supersedes #35. PR #35's branch is off an older main and deletes work
that has since landed, so this is a fresh branch off current main.
Keeps the audit log format (three event types intact), the sidebar
"Compactions (N)" pill, the history modal, and the `manual_edit` path
in `update_session_fields`. Rewrote the attribution core.

Closes #35 on merge.

### DIFF
--- a/src-tauri/src/cli/mod.rs
+++ b/src-tauri/src/cli/mod.rs
@@ -5,6 +5,7 @@ pub mod notes;
 pub mod permissions;
 pub mod prompts;
 pub mod session;
+pub mod session_attribution;
 pub mod stop;
 pub mod theme;
 pub mod ticket;
@@ -807,6 +808,50 @@ fn cmd_resume(fork: bool) -> i32 {
         }
         build_and_launch(&work_dir, &window_name, &color, Some(&session_id), &command, chrome, provider, None)
     } else {
+        // Attribution: did /compact or /clear swap in a new jsonl since the
+        // last resume? Adopt only when the chain-of-descent signal is
+        // unambiguous; otherwise prompt the user rather than silently
+        // overwriting. Runs before bumping last_resumed so the "since"
+        // filter uses the prior resume's timestamp.
+        match session_attribution::maybe_sync_session_id(&work_dir, &mut session_data) {
+            session_attribution::SessionSyncOutcome::NoChange => {}
+            session_attribution::SessionSyncOutcome::Adopted {
+                old_id,
+                new_id,
+                event,
+                ..
+            } => {
+                println!(
+                    "Detected /{event} — adopted session id {} → {} (chain-of-descent confirmed)",
+                    short_id(&old_id),
+                    short_id(&new_id),
+                );
+            }
+            session_attribution::SessionSyncOutcome::NeedsConfirmation {
+                old_id,
+                candidates,
+            } => {
+                if let Some(chosen) = prompt_session_adoption(&old_id, &candidates) {
+                    match session_attribution::adopt_candidate_confirmed(
+                        &work_dir,
+                        &mut session_data,
+                        &chosen,
+                    ) {
+                        Ok(_) => println!(
+                            "Adopted session id {} → {} (user-confirmed)",
+                            short_id(&old_id),
+                            short_id(&chosen.session_id),
+                        ),
+                        Err(e) => eprintln!("Warning: could not record adoption: {}", e),
+                    }
+                } else {
+                    println!(
+                        "Keeping stored session id {}. (Edit via the session config modal or `twapp set-session` if Claude fails to resume.)",
+                        short_id(&old_id),
+                    );
+                }
+            }
+        }
         session_id = session_data.session_id.clone();
         let command = format!("{}claude --resume {}{}", cd_prefix, session_id, chrome_flag);
         session_data.last_resumed = Some(chrono::Utc::now().to_rfc3339());
@@ -816,6 +861,72 @@ fn cmd_resume(fork: bool) -> i32 {
         }
         build_and_launch(&work_dir, &window_name, &color, Some(&session_id), &command, chrome, provider, None)
     }
+}
+
+fn short_id(id: &str) -> String {
+    if id.len() <= 8 {
+        id.to_string()
+    } else {
+        format!("{}…", &id[..8])
+    }
+}
+
+/// Interactive fallback for `cmd_resume` when attribution can't auto-resolve.
+/// Returns the candidate the user selected, or `None` to keep the stored id.
+fn prompt_session_adoption(
+    old_id: &str,
+    candidates: &[session_attribution::SessionCandidate],
+) -> Option<session_attribution::SessionCandidate> {
+    use std::io::{BufRead, Write};
+
+    // Skip the prompt when stdin isn't a tty (CI, scripted resumes) — safer
+    // to leave the stored id untouched and let the user fix it manually.
+    if !atty_stdin() {
+        return None;
+    }
+
+    eprintln!();
+    eprintln!(
+        "Claude wrote {} jsonl file(s) since the last resume of {} that don't clearly descend from it:",
+        candidates.len(),
+        short_id(old_id),
+    );
+    // Newest first.
+    let mut sorted = candidates.to_vec();
+    sorted.sort_by(|a, b| b.mtime.cmp(&a.mtime));
+    for (idx, c) in sorted.iter().enumerate() {
+        eprintln!(
+            "  [{}] {} ({})",
+            idx + 1,
+            short_id(&c.session_id),
+            c.event
+        );
+    }
+    eprint!(
+        "Pick a number to adopt that id, or press Enter to keep {}: ",
+        short_id(old_id),
+    );
+    let _ = std::io::stderr().flush();
+
+    let stdin = std::io::stdin();
+    let mut line = String::new();
+    if stdin.lock().read_line(&mut line).is_err() {
+        return None;
+    }
+    let choice = line.trim();
+    if choice.is_empty() {
+        return None;
+    }
+    let pick: usize = choice.parse().ok()?;
+    if pick == 0 || pick > sorted.len() {
+        return None;
+    }
+    Some(sorted.into_iter().nth(pick - 1).unwrap())
+}
+
+fn atty_stdin() -> bool {
+    use std::io::IsTerminal;
+    std::io::stdin().is_terminal()
 }
 
 /// Build app args, prepare instance app, and launch GUI.
--- a/src-tauri/src/cli/session_attribution.rs
+++ b/src-tauri/src/cli/session_attribution.rs
@@ -1,0 +1,800 @@
+//! Claude session-ID attribution for twapp.
+//!
+//! The problem: `~/.claude/projects/<encoded_cwd>/` is scoped per working
+//! directory, not per session. Multiple twapp sessions rooted at the same
+//! directory share that directory. Picking the newest-mtime jsonl as the
+//! replacement for a stored session id can mis-attribute a file written by
+//! session A to session B.
+//!
+//! Resolution precedence:
+//! 1. **Chain-of-descent.** A jsonl created by `/compact` carries a
+//!    `compact_boundary` line near the top whose `sessionId` field is the
+//!    parent session's id. If a candidate descends from our stored id via a
+//!    short chain of such boundaries, adopt it.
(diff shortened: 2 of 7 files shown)


piekstra (commented):
**Ship it.** This is the right fix for the misattribution concern I raised on #35 — and the test suite literally encodes the reproducer from that review as `misattribution_guard_does_not_adopt_sibling_sessions_compact`.

## Policy scan

Clean on the extended regex (`schwab|scalper|auto.?sell|auto.?buy|pair.?seq|break.?even|bs.?mark|ufo|conl|msfu|options_scalper|walk.limit.*schwab|tos.?ws|BP.?formula|comm.*\$0\.65`). Test session ids (`A1`/`A2`/`A3`/`B1`/`C1`) and fake `projA`/`shared`/`projChain`/`projSync` directory names are the right kind of generic.

## Correctness — the four resolution rules

**1. Chain-of-descent (primary signal).** `descends_from` walks the candidate's `compact_boundary` parent chain up to 8 hops:
- Cycle guard via `HashSet<String> seen` — returns false on revisit.
- Terminates cleanly on: parent missing (`compact_parent_session_id: None`), parent file not readable, `stored_id` hit, or depth exceeded.
- Correct single-hop: on iteration 1, if `parent_id == stored_id`, returns true immediately. A1→A2 resolves in one hop.
- Correct two-hop: the `chain_of_two_compactions_still_resolves` test exercises A1→A2→A3. ✓
- `read_jsonl_head` only sets `compact_parent_session_id` when the `sessionId` on the `compact_boundary` line **differs** from the file stem. Right call — if they match, the line describes the same session, not a parent link.
- Upgrade from PR #35's substring match: `read_jsonl_head` now parses the compact_boundary line as JSON (`serde_json::from_str::<serde_json::Value>`) before reading `sessionId`. Much less brittle if Claude ever reformats the field.
- Termination order inside `read_jsonl_head` is correct: a `compact_boundary` hit sets everything + breaks; a `summary` / `isCompactSummary` hit sets `looks_compacted = true` but continues scanning for a later `compact_boundary`.

**2. Cross-session exclusion.** `collect_other_claimed_ids` scans the global `work_directory` for sibling `.twapp-session.json` files and pulls their `session_id` + `codex_session_id` into the exclusion set. Self-exclusion uses `canonicalize` on both paths (falling back to raw on failure) — safe either way because `find_session_candidates` also explicitly skips `head.session_id == stored_id`. Belt-and-suspenders as the PR body says.
- Edge case: if the user has multiple work directory trees (e.g. `~/work1` and `~/work2`), only the globally-configured one is scanned. The other tree's sessions won't enter the exclusion set. But the chain-of-descent check is the primary signal, so this is an acceptable gap; documented behavior.

**3. `last_resumed` narrowing.** `parse_rfc3339_as_systemtime` + mtime comparison excludes jsonls older than the previous successful resume. Tested by `last_resumed_at_narrowing_excludes_old_jsonls`. Correct ordering: `cmd_resume` calls `maybe_sync_session_id` **before** bumping `last_resumed`, so the attribution window uses the *previous* resume's timestamp. Explicit comment in the GUI path too. ✓

**4. Interactive fallback.** `SessionSyncOutcome::NeedsConfirmation` surfaces candidates without writing. CLI's `prompt_session_adoption`:
- TTY check via `std::io::stdin().is_terminal()` — non-tty (CI, scripted resumes) returns None, leaving the stored id intact. Right default.
- Sorted newest-first; 1-indexed pick; guards against `pick == 0` and out-of-bounds.
- Empty input keeps the stored id — users can no-op safely.

GUI `launch_session` deliberately **does not** auto-adopt on `NeedsConfirmation` — the PR body says "users still reach the same code path via the session config modal, which logs a `manual_edit` event". That path still works because `update_session_fields` snapshots `prior_claude_id` + `prior_codex_id` before the write and logs on real change. Correct belt-and-suspenders.

## Audit-log extension

Two optional fields on `SessionHistoryEvent`:
- `ambiguous: Option<bool>` — set on user-confirmed adoptions.
- `reason: Option<String>` — `"descent_chain"` / `"user_confirmed"` / `"manual_edit"`.

Both use `#[serde(default, skip_serializing_if = "Option::is_none")]` — forward-compatible if any `.twapp-session-history.json` from PR #35's previous incarnation exists on disk (read path tolerates absent fields; write path omits them when None). Legacy audit entries deserialize cleanly; new entries serialize tightly. ✓

The ambiguous badge in the GUI modal (gray, tooltipped `"Adopted without a chain-of-descent signal — user-confirmed"`) gives the user a post-hoc cue about which adoptions were automatic vs human-confirmed. Nice touch.

## Test coverage

9 tests, all in `session_attribution::tests`. Coverage matrix:

| Test | Covers |
|---|---|
| `happy_path_chain_of_descent_adopts_compacted_id` | 1-hop direct adoption |
| **`misattribution_guard_does_not_adopt_sibling_sessions_compact`** | **My reproducer from #35 — exact** |
| `cleared_unrelated_session_returns_needs_confirmation_not_auto_adopt` | `/clear` → no auto-adopt |
| `cross_session_exclusion_skips_other_claimed_ids` | Excluded set drops candidates |
| `last_resumed_at_narrowing_excludes_old_jsonls` | `since` filter excludes ancient |
| `chain_of_two_compactions_still_resolves` | Multi-hop descent |
| `maybe_sync_writes_session_and_history_on_adoption` | Write path happy case |
| `maybe_sync_returns_needs_confirmation_on_cleared_unrelated` | Write path NeedsConfirmation no-op |
| `adopt_candidate_confirmed_flags_ambiguous_in_audit` | Ambiguous + user_confirmed labeling |

Test infrastructure notes:
- `HOME_LOCK: Mutex<()>` serializes the 9 tests that mutate `HOME` env var — correct, because cargo parallelizes within a binary.
- `bump_mtime()` sleeps 50ms to ensure strictly-increasing mtimes. Fine for APFS / ext4 / btrfs / zfs (all nanosecond-precision). Would be flaky on older HFS+ (1s resolution), but HFS+ hasn't been macOS default since 10.13. Acceptable.
- `HomeGuard` cleans up tmp directories on Drop; tests hermetic.

One small nit on the test infrastructure: the `unsafe { std::env::set_var(...) }` calls are required on Rust 1.82+ (where `set_var` is marked unsafe due to thread-safety). The mutex makes them safe here — good.

## Non-blocking nits for a future follow-up

- **`append_history` non-atomicity.** Read → push → write is still non-atomic; concurrent writers (a racing `cmd_resume` vs GUI `update_session_fields`) can lose events. Same observation I flagged on #35 — low likelihood on a single-user workstation, not worth a lockfile today, but worth one sentence in the module docstring acknowledging the assumption.
- **No `append_history` size cap.** A session that gets compacted hundreds of times grows the history file unboundedly. Not urgent; rotate-at-N-entries is a reasonable follow-up.
- **`classify_session_change_at` heuristic**. Still relies on a substring scan of the first 20 lines for `type:summary` / `isCompactSummary:true` / `subtype:compact_boundary`. Now that `read_jsonl_head` parses JSON for the parent id, consider upgrading `classify_session_change_at` to the same JSON parse — would mean one more dependency on Claude's exact field naming but exclude false positives from stray body text.
- **`collect_other_claimed_ids` scan root.** Uses global `work_directory` or falls back to `current_work_dir`'s parent. Users with multiple separate project trees outside the global root won't have their sessions' ids in the exclusion set. The chain-of-descent primary signal covers the common case, so this is fine; worth a one-liner comment naming this gap so future reviewers don't think it's a bug.
- **Codex session ids.** `update_session_fields` detects changes on both `session_id` and `codex_session_id`. Tie-breaker logic only logs one of the two on a simultaneous change — probably never happens in practice, but a note saying "simultaneous change picks claude; if that's wrong, log separately" would make future maintenance easier.

## Summary

Replaces PR #35's mtime-only heuristic with a chain-of-descent signal backed by cross-session exclusion, `last_resumed` narrowing, and an interactive fallback. The GUI deliberately stays conservative and leaves ambiguous cases to the user. Audit log gains `ambiguous` + `reason` so automatic adoption is distinguishable from manual edits.

Tests encode the exact misattribution reproducer from my #35 review; CI is green (`PR Check` succeeded at 21:54:21Z). The UNSTABLE `mergeStateStatus` is from absence of approving review, not a check failure.

**Ship it.** Closes #35 correctly. Happy to see this one land.

PR #43, opened 2026-04-20
feat: session metadata — role + provenance fields

piekstra:
## Summary
- Adds `role` and `provenance` fields to `.twapp-session.json` (both `Option<String>`, serde-default, skip-serializing-if-none, so legacy files load unchanged).
- `twapp work` gains `--role <ROLE>`, `--spawned`, and `--provenance <VAL>`. Empty `--role` / `--provenance` error out. Precedence: `--provenance` > `--spawned` > `--from-file` implies spawned > default `"user"`.
- `twapp sessions` output now includes a Role column: `[impl] spawned` format (4-char role tag, `spawned` shown only for spawned-provenance). `-` when neither set.
- GUI `create_and_launch_session` stamps `provenance="user"` on GUI-initiated sessions; the fork path inherits role/provenance from the parent.

## Design notes
- Metadata plumbing only — no React/Tauri UI changes (that's `twapp-ui-role-badge`, a later PR).
- `role` is free-form string at this layer; the UI will validate against the archetype list.
- `--from-file` auto-implies `provenance="spawned"` so agent-launch scripts don't need to pass `--spawned` on every invocation. `--provenance user` overrides for the edge case.

## Test plan
- [x] `cargo test --lib` — all 60 unit tests pass, including 17 new ones.
- [x] `cargo check` clean.
- [x] `twapp work --help` shows the new flags with human-readable descriptions.
- [ ] Reviewer: verify legacy `.twapp-session.json` (without role/provenance) still loads and round-trips without spontaneously gaining fields.
- [ ] Reviewer: verify sessions list layout still readable on narrow terminals (added ~16 chars).

## Coordination
- Coordinates with `twapp-coordinator-launch-cli` (parallel): that PR consumes this `--role` flag. If they land first, they'll add a follow-up; if this lands first, they adopt directly.

Refs briefing: `twapp-session-role-metadata`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,7 +561,7 @@ defaults:
 
 | File | Location | Purpose |
 |------|----------|---------|
-| `.twapp-session.json` | Working dir | Session metadata, provider handles, fork ancestry |
+| `.twapp-session.json` | Working dir | Session metadata, provider handles, fork ancestry, role + provenance |
 | `.twapp-notes-{name}.json` | Working dir | Session notes |
 | `.twapp-prompts-{name}.json` | Working dir | Project-scoped quick prompts |
 | `.twapp-ticket.json` | Working dir | Linked ticket metadata |
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -10,7 +10,7 @@ The long-running orchestration loop that wraps around many agent launches.
 
 This skill teaches a Claude instance how to act as a **coordinator** — managing the lifecycle of two or more twapp-hosted worker agents, reviewing their work, merging their PRs, and keeping a shared mailbox tidy.
 
-For launching a single agent, use the [`spawn-agent`](../spawn-agent/SKILL.md) skill (coming in a related PR if not yet present). This skill assumes you already know how to spawn one agent; it covers what to do when you have many.
+For launching a single agent, use the [`spawn-agent`](../spawn-agent/SKILL.md) skill. This skill assumes you already know how to spawn one agent; it covers what to do when you have many.
 
 ## 1. When to use
 
@@ -54,7 +54,11 @@ offboard shape, hard rules.>
 sync via mailbox before touching shared files.>
 
 ## Worktree
-<Copy-paste `git worktree add` block + `.claude/settings.local.json` seed.>
+<Copy-paste `git worktree add` block + `.claude/settings.local.json` seed +
+the `twapp work --name <handle> --role <role> --from-file <briefing>` command
+the coordinator will run to spawn this worker. `--role` is the §13 archetype
+for this worker; `--from-file` auto-tags the session as agent-spawned so UI
+and `twapp sessions` can distinguish it from human-driven sessions.>
 
 ## Domain facts (if units / money / safety are involved)
 <Explicit statements of the domain model the worker must internalize before
@@ -249,8 +253,8 @@ The worker reads it on the next poll, archives, and stops. If the worker's `/loo
 ### Coordinator cleanup post-offboard
 
 ```
-twapp stop --name <handle>                 # graceful host shutdown (coming in a related PR)
-kill -TERM $(pgrep -f "twapp --name <handle>")   # fallback today
+twapp stop --name <handle>                 # graceful host shutdown
+twapp stop --name <handle> --force         # SIGKILL fallback if SIGTERM doesn't land
 git worktree remove <path>                 # optional, keeps repo lean
 ```
 
@@ -368,10 +372,17 @@ Each briefing follows §2. Worker A owns `src/parser.ts`; worker B owns `src/ren
 
 ### 2. Coordinator spawns both workers
 
-Using the `spawn-agent` skill's file-reference pattern (see that skill for the full command), one call per worker. Then:
+Using the `spawn-agent` skill's file-reference pattern, one call per worker. Tag each with its §13 role so `twapp sessions` and later UI can distinguish implementer workers from the coordinator / reviewer:
 
 ```
-twapp sessions                                   # both instances appear
+twapp work --name worker-a --role implementer --from-file /tmp/collab/briefings/worker-a.md
+twapp work --name worker-b --role implementer --from-file /tmp/collab/briefings/worker-b.md
+```
+
+`--from-file` auto-sets `provenance=spawned` — no need to pass `--spawned` alongside it. Then:
+
+```
+twapp sessions                                   # both instances appear, role column shows [impl] spawned
 ls /tmp/collab/mailbox/inbox/ | grep -E "worker-a|worker-b"
 ```
 
@@ -409,13 +420,13 @@ gh pr merge <N> --merge --delete-branch
 Each worker posts its offboard message (§7a) and archives its read mail. The coordinator acknowledges in the mailbox, then:
 
 ```
-kill -TERM $(pgrep -f "twapp --name worker-a")
-kill -TERM $(pgrep -f "twapp --name worker-b")
+twapp stop --name worker-a
+twapp stop --name worker-b
 git worktree remove /path/to/worker-a-worktree
 git worktree remove /path/to/worker-b-worktree
 ```
 
-(When `twapp stop --name <handle>` lands in a related PR, prefer it over `kill -TERM`.)
+`twapp stop` SIGTERMs the host and the child Claude; add `--force` to escalate to SIGKILL if the host didn't exit within ~3s.
 
 ### 7. Coordinator stays alive
 
--- a/skills/spawn-agent/SKILL.md
+++ b/skills/spawn-agent/SKILL.md
@@ -26,13 +26,13 @@ Inline prompts in `--run` are shell-fragile. A long prompt with curly quotes, `-
 
 **Rule: write the briefing to a markdown file first, then reference it.**
 
-Preferred (once the `--from-file` flag ships — see [Coming soon](#coming-soon)):
+Preferred:
 
 ```bash
 twapp work --name my-agent --from-file /absolute/path/to/briefing.md
 ```
 
-Manual equivalent that works on any current twapp version:
+Manual equivalent on a twapp build that predates `--from-file`:
 
 ```bash
 twapp work --name my-agent \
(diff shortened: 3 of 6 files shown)


piekstra (commented):
Review — plumbing is sound, tests are thorough, but there's one CLI/GUI behavioral divergence worth resolving before ship-it.

## Policy scan

Clean on the extended regex. Test examples (`coordinator`, `implementer`, `audit`, `worker`) are drawn from the already-landed `agent-coordinator` skill §13 vocabulary — right source.

## Answering your three specific questions

### Q1 — Precedence for `--provenance` vs `--spawned`

Reasonable. The rule (`--provenance` wins; errors if `--spawned + --provenance user` disagree; accepts `--provenance spawned + --spawned` as redundant-but-fine) is the right ergonomics:
- `resolve_provenance(Some("user"), true, false)` → error ✓
- `resolve_provenance(Some("spawned"), true, false)` → `"spawned"` ✓
- `resolve_provenance(Some("user"), false, true)` → `"user"` — explicit override of `--from-file`'s implicit default ✓

One thing to note (non-blocking): `resolve_provenance` accepts **any** non-empty string, so `--provenance spawnd` (typo) silently ships with the typo. The design doc (`docs/designs/agent-aware-ui.md` §3.1) defines exactly two values (`"user"` / `"spawned"`); your `format_role_cell` renders only `"spawned"` specially and silently treats everything else as non-spawned. Worth one of:
1. Allow-list validation in `resolve_provenance` (reject anything other than `user` / `spawned`).
2. A doc-comment saying unknown values are forward-compat (future: `"imported"`, etc.) but currently render as "not spawned".
Either is fine; pick one so a future reader isn't guessing.

### Q2 — Fork inherits role+provenance

**This is the one real concern.** The CLI and GUI fork paths disagree:

- **`cmd_resume` with `fork=true`** (`src-tauri/src/cli/mod.rs:908-910`):
  ```rust
  role: session_data.role.clone(),
  provenance: session_data.provenance.clone(),
  ```
  Inherits from parent. ✓ Matches your stated intent.

- **`gui/sessions.rs::fork_session`** (two places — `:1492-1493` and `:1536-1537`):
  ```rust
  role: None,
  provenance: None,
  ```
  Does **not** inherit; sets both to None. Diverges from the CLI.

For plumbing-only, the two code paths should behave identically. A user who forks via `twapp resume --fork` keeps their role/provenance; a user who forks via the GUI fork button loses both. That's surprising and undermines the fork-inherits-from-parent contract the CLI test (`resolve_provenance` tests don't cover fork explicitly, but `cmd_resume` clones the fields) implies.

**Suggested fix:** in `fork_session`, clone from the parent `SessionData` the same way CLI does. Both code paths already read the parent session file earlier in their flow, so the fields are in scope.

(If you specifically want the GUI fork to *reset* role/provenance for some design reason, that belongs in the agent-aware-ui implementation PR, not this plumbing PR — and needs a design-doc update since `agent-aware-ui.md` §3.1 doesn't say forking resets.)

### Q3 — Sessions-list column width

Readable. The 16-char Role column absorbs `[impl] spawned` (14 chars), `[coordinator]` truncated to `[coor]` (6 chars). The earlier output was already ~100 chars, so wrap behavior on narrow terminals isn't meaningfully worse. `format_role_cell` correctly returns `-` for both-None so the column isn't visually noisy for single-session users. ✓

## Other correctness checks

- **Serde roundtrip.** `#[serde(default, skip_serializing_if = "Option::is_none")]` on both fields. Legacy files round-trip cleanly (`legacy_session_file_deserializes` test proves this); new fields write only when set. No `.twapp-session.json` on disk today is touched unless the user re-writes it with a role/provenance set.
- **`import_sessions` sets both to `None`.** Arguably imports are user-initiated and should get `Some("user")`, but None is defensible as "unknown / legacy shape". Non-blocking — if `format_role_cell` already treats `None` as `-`, the displayed result is the same.
- **`create_session_core` signature**: two new trailing params (`role: Option<String>`, `provenance: Option<String>`). GUI callers updated to pass `None, Some("user")` (`create_and_launch_session`) — correct, "Launch Session" click is user-initiated by construction.
- **`validate_role` trims but doesn't canonicalize.** `--role Coordinator` and `--role coordinator` produce different strings. Design doc §3.2 says role tokens are `kebab-case`; PR accepts any string. Non-blocking — the consumer UI would canonicalize on read if it wants normalization, and keeping this layer free-form matches the PR description ("free-form — not enum-enforced at this layer; UI does any validation").

## Test coverage

17 tests, all of them correctly scoped:

| Test | Verifies |
|---|---|
| `session_serde_roundtrip_with_role_and_provenance` | Round-trip both present |
| `session_serde_roundtrip_omits_absent_role_and_provenance` | `skip_serializing_if` actually skips |
| `write_then_list_sessions_preserves_role_and_provenance` | Disk round-trip |
| `legacy_session_file_deserializes` | Forward-compat with pre-PR session files |
| `empty_role_errors` / `whitespace_role_errors` / `populated_role_passes_through_verbatim` / `missing_role_is_none` | `validate_role` contract |
| `direct_work_defaults_to_user` / `spawned_flag_sets_spawned` / `from_file_implies_spawned_provenance` | Implicit defaults |
| `explicit_provenance_user_overrides_from_file_default` / `explicit_provenance_verbatim_passes_through` | Explicit overrides |
| `empty_provenance_errors` / `provenance_and_spawned_disagreement_errors` / `provenance_spawned_plus_spawned_flag_is_fine` | Error + redundant-but-consistent cases |
| `role_cell_*` (5 tests) | Display formatting |

Missing (follow-up, not blocking this PR):
- A test that exercises the actual CLI→disk path with `--role` and `--provenance` flags set. Current coverage is unit-level on `validate_role` / `resolve_provenance` / `format_role_cell` and round-trip through `SessionData`, but no test invokes `cmd_work` end-to-end with these flags. That's fine — integration tests for the CLI surface would be a larger scaffolding. The unit coverage is dense enough.
- A GUI fork inheritance test (once Q2 is resolved).

## Summary

- Q1: precedence rule is reasonable; consider allow-list validation for `--provenance` (or just document the forward-compat intent).
- **Q2: CLI/GUI fork behavior diverges — CLI inherits, GUI resets to None. Either fix the GUI to inherit (matching CLI), or document the reset decision in the design doc. Blocker for "plumbing only" framing.**
- Q3: column width is fine.

Not flipping to ship-it yet, pending the GUI fork inheritance decision. Happy to re-review as soon as that's addressed — the rest of the PR is clean, the serde compat is careful, and the test suite pins the precedence rules before the agent-aware-ui PRs build on them.

piekstra (commented):
**Re-review at tip 68a3b844 — all three concerns from my prior review addressed. Ship it pending rebase.**

## Policy scan

Clean on the extended regex.

## What got fixed

### Q2 (the real concern): GUI fork inheritance ✓

`gui/sessions.rs::fork_session` now inherits `(role, provenance)` from the parent session via a new `read_fork_inherited_metadata` helper. Both branches (the imported-fork path and the else-branch fresh-fork path) thread `parent_role` / `parent_provenance` into the new `SessionData`. The docstring on the call site explains the *why* directly:

> A fork is a derivative of the parent's context, not a fresh launch — resetting these would drop the agent tag on every fork.

Right framing. CLI `twapp resume --fork` and the GUI fork button now agree.

The new `fork_inheritance_tests::inherits_role_and_provenance_from_parent_session_file` test exercises the helper. Worth a small nit (non-blocking): the test name implies a single-case test, but the commit message mentioned "present / missing / legacy parent cases" — if those are separate tests further in the file, ✓; if they're folded into one assertion, it'd read better as three tests. I can only see the head of the test in the diff.

### Q1: forward-compat doc on `resolve_provenance` ✓

Per the commit message: "Doc-comment on `resolve_provenance` makes the forward-compat intent explicit: non-user/spawned values are accepted verbatim but render as not-spawned in UI. Leaves room for future provenance kinds without a schema bump." Right resolution — doesn't constrain the schema; documents the intent so a future `--provenance imported` reader knows what to expect.

### Skill-doc updates ✓

Both skill files get the agent-aware updates the coordinator's directive at 22:10:00Z asked for, and they go beyond the minimum — examples A and B in spawn-agent now use `--role implementer --from-file`, agent-coordinator's §13 end-to-end example uses `twapp work --role implementer --from-file` instead of the manual `--run` form, and the offboard block in §13 uses `twapp stop` instead of the stale `kill -TERM` fallback. Stale "coming in a related PR" notes whose items shipped (#36's spawn-agent existence, #38's `twapp stop` and `--from-file`) are removed.

The new spawn-agent "Role + provenance" section is well-shaped:
- Names which flag does what (`--role`, `--from-file`, `--spawned`, `--provenance`).
- Cross-links to agent-coordinator §12 for the role archetype list.
- Shows the `twapp sessions` row format with `[impl] spawned`.
- Calls out the no-flag default (`-` in the column), explicitly preserving non-disruption for pre-existing workflows.

## ⚠️ Blocker: `mergeStateStatus: DIRTY`

The PR has a merge conflict against `main`. Almost certainly with #47 (`docs(skills): mandate headless operation via /loop`), which merged earlier and also touches both `skills/spawn-agent/SKILL.md` and `skills/agent-coordinator/SKILL.md`. The two PRs edit overlapping regions:

- **#47 added** `Headless operation` sections to both skills + a `/loop` mandate line in agent-coordinator §2 Protocol template.
- **#43 modifies** the agent-coordinator §2 Worktree template + spawn-agent's Examples A/B + spawn-agent's "Coming soon" section + agent-coordinator §13 end-to-end example.

The conflicts shouldn't be semantically fraught — neither PR drops or replaces the other's content — but the resolution needs careful merging:

1. Rebase against `main` (which now has #47's headless sections).
2. Keep the new "Headless operation" sections from #47.
3. Apply #43's template/example/coming-soon edits on top.
4. Re-run `cargo test` and re-verify `twapp sessions` example output reads naturally next to the `/loop` mandate line in the §2 template.

Once the rebase is in and CI is green, this is ship-it.

## Other notes (non-blocking)

- **`create_session_core` signature** now takes `role` and `provenance` as the 4th and last positional args. GUI `create_and_launch_session` correctly stamps `Some("user".to_string())`. This conflicts with #44's planned `create_session_core` signature (which adds a different `role` param at a different position) — whichever lands second will need to thread #43's `provenance` through #44's `cmd_launch` call (or vice versa). PRs #44 and #43 already coordinated on this in their hellos.
- **`read_fork_inherited_metadata` returns `(None, None)` on read failure.** Documented in the helper's docstring as "treated as a plain session". Defensive — a fork off a parent session file that's been partially written won't crash. Good.
- **The §2 Worktree template now embeds the spawn command** including `--role` and `--from-file`. The change is *inside* the angle-bracket template description block (which is the literal text coordinators copy-paste into briefings). That's intentional — coordinators will carry the new spawn syntax forward without re-thinking it. ✓

## Summary

All three of my prior concerns addressed cleanly:
1. GUI/CLI fork divergence: fixed with a real helper + tests. ✓
2. Forward-compat doc on `resolve_provenance`: present per commit message. ✓
3. Skill-doc updates per coordinator's directive: comprehensive, including bonus stale-content cleanup. ✓

**Ship it once the rebase resolves the conflict with #47** (and CI re-runs green). The fix is right; the merge state just needs untangling.

piekstra (commented):
Quick re-confirmation at tip `209f3236`: rebase resolved cleanly per the rebase commit message ("agent-coordinator SKILL.md auto-merged cleanly — #47's headless operation block + this branch's `--role` guidance live side-by-side"), `cargo test --lib` reports 83 passing, CI is green. The skill files now carry both #47's headless-op section and this branch's role/provenance flag guidance side-by-side without conflict, exactly as I'd hoped.

**Ship it** (this is the unconditional ship-it that my prior comment was conditional on).

PR #41, opened 2026-04-20
feat: msg CLI scaffolding (send / broadcast / fetch) — design PR-1

piekstra:
## Summary

- Adds `twapp msg {send,broadcast,fetch}` — a thin CLI wrapper over the shared filesystem mailbox described in `docs/designs/agent-messaging.md`. PR-1 of the 7-PR plan in that doc's §5.
- Emits fenced-YAML frontmatter (`id`, `from`, `to[]`, optional `cc[]`, `priority`, `subject`, `thread`, `in_reply_to`, `ts`) into the existing flat `<mailbox>/inbox/<YYYYMMDDTHHMMSSZ>-<id6>.md` — zero structural change. Directory split, read cursors, priority lanes, presence, channels, and archive rotation land in PRs 2-7.
- Reads both the new fenced shape **and** the bare legacy `from:` / `to:` / `re:` files so day-one upgrades don't break inboxes where the audit corpus is 84% bare.

## What's in the wire format

```yaml
---
id: <20-char Crockford base32 — 100 bits random, stable, 8-26 chars per §2.2>
from: <handle>
to: [<handle>, ...]          # array; [all] for broadcast; ["channel:<name>"] for channels
cc: [<handle>, ...]          # optional; omitted when empty
priority: routine            # routine | urgent | blocker; default routine
subject: "<short subject>"   # optional
thread: <id>                 # omitted on new threads (self-id is implicit)
in_reply_to: <id>            # omitted on thread root
ts: YYYYMMDDTHHMMSSZ         # duplicates filename ts — survives mtime clobber
---

<markdown body>
```

IDs are pure-random Crockford base32 (not time-prefixed ULIDs) so the 6-char filename slice stays unique even when two messages land in the same second.

## Configuration

- `TWAPP_MAILBOX_DIR` — preferred; points directly at the mailbox root.
- `TWAPP_SHARED_DIR` — fallback; mailbox is `<shared>/mailbox/`.
- Neither set → clear error naming both env vars.
- `--from` falls back to `.twapp-session.json` `name` in the cwd; missing on both → error.
- `fetch` without `--for` defaults to the current session handle; `--all` to list everything.

## Test plan

- [x] `cargo test --lib` — 39 tests green (26 new in `cli::msg`).
- [x] send writes a well-formed fenced-frontmatter file with the required fields.
- [x] broadcast sets `to: [all]`; `--channel foo` switches to `to: [channel:foo]`.
- [x] fetch `--for <h>` returns direct + broadcast + cc messages.
[...]

PR #44, opened 2026-04-20
feat: twapp coordinator launch | claim subcommands

piekstra:
## Summary
- Adds `twapp coordinator launch [--briefing <p>] [--name <n>] [--cwd <d>] [--shared-dir <d>]` — spawns a fresh session wired as coordinator; internally equivalent to `twapp work --name <n> --role coordinator --from-file <briefing>` with mailbox plumbing.
- Adds `twapp coordinator claim [--name <n>] [--force]` — flips `.twapp-session.json:role` in place. Refuses to stomp on a populated non-coordinator role without `--force`.
- Ships `templates/coordinator-bootstrap.md` — generic bundled bootstrap (63 lines) that points new coordinators at `skills/agent-coordinator/SKILL.md`, `docs/designs/agent-messaging.md`, and the mailbox protocol. Embedded via `include_str!` so `twapp coordinator launch` writes it next to the new session at `.twapp-coordinator-bootstrap.md`; users override via `--briefing <path>`.
- README gains a "Running a coordinator" section covering both commands and mailbox precedence.

## Shared-dir handling (per briefing)
1. `--shared-dir <dir>` → export `TWAPP_MAILBOX_DIR=<dir>` for the spawned session.
2. Otherwise inherit `$TWAPP_MAILBOX_DIR` if set.
3. Otherwise reuse `<cwd>/mailbox/` if its `inbox/` already exists.
4. Otherwise create `<cwd>/collab/mailbox/` and print the path to stderr.

## Guardrails
- `launch` refuses to overwrite an existing session at the target directory (error directs the user to `twapp coordinator claim` or `twapp stop`).
- `claim` refuses to clobber an existing non-coordinator role without `--force` (is a no-op on existing `role=coordinator`).

## Dependency
Adds a minimal forward-compat shim for `role: Option<String>` on `SessionData` plus `--role <ROLE>` on `twapp work`, so this PR compiles and ships standalone if it lands first. The full role-metadata surface (provenance, sessions-list formatting, GUI stamping, etc.) lives in #43 (`feature/session-role-metadata`). Merge order is flexible:

- If #43 lands first, this PR's shim collapses into its schema at rebase time.
- If this PR lands first, #43 extends the same `role` field with provenance + formatting.

Coordinated with `twapp-session-role-metadata` via mailbox (hello thread).

## Test plan
- [x] `cargo test --lib` — 46/46 pass, 7 new coordinator tests.
- [x] Required briefing tests all covered:
  - `launch_writes_role_coordinator` — verifies the `.twapp-session.json:role` schema contract round-trips end-to-end via `session::write_session` / `read_session`, plus that the launch call site uses the `"coordinator"` literal (compile-asserted constant).
  - `launch_without_briefing_uses_default_template` — bundled template materializes to an absolute on-disk path and matches the shipped content.
  - `launch_refuses_existing_coordinator` — guardrail predicate refuses overwrite.
  - `claim_flips_role_in_place` — role-unset session is flipped to `coordinator`.
  - `claim_refuses_overwriting_role_without_force` — refused with exit 2, role preserved; `--force` succeeds.
- [x] Extra: mailbox resolution precedence tests (`--shared-dir` wins; fallback creates `collab/mailbox/`).
- [x] Manual smoke: `twapp coordinator claim` on a hand-crafted session file writes `"role": "coordinator"`; a second claim is a no-op.
- [x] Manual smoke: `twapp coordinator launch --briefing /nonexistent.md` fails fast with exit 2 before touching GUI.
- [x] `twapp coordinator --help`, `launch --help`, `claim --help` render cleanly.
- [ ] Reviewer: end-to-end launch through the GUI (requires installed app bundle) to confirm `TWAPP_MAILBOX_DIR` lands in the spawned claude's env.

## Out of scope
- Presence file (`presence/<handle>.json`) — that's messaging PR-5.
- GUI "Become coordinator" button.
- Enforcement of one-coordinator-at-a-time beyond the name collision check — presence-based enforcement lands with messaging PR-5.
- Any role other than `coordinator`.

Refs briefing: `twapp-coordinator-launch-cli`. Depends on #43.

### DIFF
--- a/README.md
+++ b/README.md
@@ -473,37 +473,39 @@ automating spawns.
 
 ### Bootstrapping a coordinator
 
-> **Coming in [#44](https://github.com/piekstra/twapp/pull/44).** That
-> PR adds `twapp coordinator launch` and `twapp coordinator claim`,
-> which wrap the moves below into a one-liner and ship a bundled
-> bootstrap briefing pointing new coordinators at
-> [`skills/agent-coordinator/SKILL.md`](skills/agent-coordinator/SKILL.md).
-
-A coordinator today is just a regular `twapp work` session opened in
-whatever directory holds your mailbox, with the
-[`agent-coordinator`](skills/agent-coordinator/SKILL.md) skill in scope.
+Use `twapp coordinator launch` to stand up a coordinator session —
+`.twapp-session.json` is stamped with `role: "coordinator"`, a bundled
+bootstrap briefing is materialized next to the session, and
+`TWAPP_MAILBOX_DIR` is plumbed through so `twapp msg` / `twapp
+coordinator claim` in the new terminal see the same mailbox without
+ceremony.
 
 ```bash
-# Start a coordinator session (pre-#44).
-export TWAPP_MAILBOX_DIR=$HOME/collab/mailbox
-mkdir -p "$TWAPP_MAILBOX_DIR/inbox" "$TWAPP_MAILBOX_DIR/archive"
-twapp work --name coordinator --cwd "$HOME/collab"
-```
-
-Once #44 lands:
-
-```bash
-twapp coordinator launch                       # uses a bundled bootstrap briefing
+twapp coordinator launch                       # uses the bundled bootstrap briefing
 twapp coordinator launch --briefing /path/to/bootstrap.md
 twapp coordinator launch --shared-dir ~/collab/mailbox
 twapp coordinator claim                        # flip an existing session's role in place
 ```
 
-The bundled bootstrap briefing points the new coordinator at the
-`agent-coordinator` skill, the messaging design doc, and the mailbox
-protocol, so you don't have to hand-roll a prompt when you stand up a
-fleet. Override it with `--briefing <path>` when you want a
-project-specific bootstrap.
+`launch` refuses to overwrite an existing session at the target
+directory — use `twapp coordinator claim` to take over in place, or
+`twapp stop` the old session first. `claim` rewrites only the `role`
+field in `.twapp-session.json` and refuses to overwrite an existing
+non-coordinator role without `--force`.
+
+`--shared-dir` precedence for the mailbox: the flag wins; otherwise
+`TWAPP_MAILBOX_DIR` is inherited from the parent env; otherwise
+`./mailbox/` under the coordinator cwd is reused if it already exists;
+otherwise `./collab/mailbox/` is created under the cwd and the path is
+printed to stderr.
+
+The bundled bootstrap
+([`templates/coordinator-bootstrap.md`](templates/coordinator-bootstrap.md))
+points the new coordinator at the
+[`agent-coordinator`](skills/agent-coordinator/SKILL.md) skill, the
+messaging design doc, and the mailbox protocol, so you don't have to
+hand-roll a prompt when you stand up a fleet. Override it with
+`--briefing <path>` when you want a project-specific bootstrap.
 
 ### Messaging between sessions
 
@@ -765,6 +767,8 @@ burns iteration.
 | `twapp msg claim <lane-id> [--note <s>]` | Atomically claim a shared lane (PR, audit, backlog item); exit 1 if already claimed |
 | `twapp msg release <lane-id> [--note <s>]` | Release a lane you own; writes `released.json` and broadcasts the release |
 | `twapp msg claim --list [--lane-prefix <p>]` | List all active (unreleased, unstale) claims; `--format json` for machine-readable |
+| `twapp coordinator launch [--briefing <p>] [--name <n>] [--shared-dir <d>]` | Spawn a fresh session wired as coordinator (writes `role: "coordinator"`) |
+| `twapp coordinator claim [--name <n>] [--force]` | Re-tag an existing session's role to `coordinator` |
 | `twapp install-gui <binary>` | Install or update the app bundle |
 | `twapp setup-cert` | Create code signing certificate |
 | `twapp dev-reload --pid <pid>` | Rebuild and relaunch (dev workflow) |
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -26,6 +26,42 @@ Invoke this skill when the caller is acting as a coordinator — i.e., spawning 
 
 Do **not** invoke this skill for single one-shot agent launches — `spawn-agent` is the right primitive for those.
 
+## 1b. Bootstrapping a coordinator
+
+Use `twapp coordinator launch` to spawn the coordinator session itself —
+it's the canonical entry point, preferred over the generic
+`twapp work --name coordinator --from-file …` invocation.
+
+```bash
+# Default: writes role=coordinator and materializes the bundled bootstrap
+# at <cwd>/.twapp-coordinator-bootstrap.md, then spawns a twapp session.
+twapp coordinator launch
+
+# Point at a project-specific briefing and a shared mailbox root.
+twapp coordinator launch \
+  --briefing /absolute/path/to/coord-briefing.md \
+  --shared-dir /tmp/collab \
+  --name ux-coord
+
+# Already running as a generic session? Re-tag it in place instead.
+twapp coordinator claim                 # claims the session in the current cwd
+twapp coordinator claim --name worker-a --force  # overwrites an existing non-coord role
+```
+
+`launch` sets `role: "coordinator"` in `.twapp-session.json` and plumbs
+`TWAPP_MAILBOX_DIR` through to the spawned claude so downstream
+`twapp msg` / `twapp coordinator claim` invocations see the same mailbox
+without ceremony. The bundled default briefing is
+[`templates/coordinator-bootstrap.md`](../../templates/coordinator-bootstrap.md) —
+it tells the new coordinator to read this skill, register in
+`handle.txt`, post a hello, and start `/loop` polling. Override it with
+`--briefing <path>` whenever project-specific scope needs to ride along.
+
+`claim` exists for the case where you're already in a running session and
+realize mid-flight that you are the coordinator — it only rewrites the
+`role` field, leaving everything else intact. Refuses to stomp on an
+existing non-coordinator role without `--force`.
+
 ## 2. Briefing structure
 
 Each worker receives a markdown file on disk and reads it via the `spawn-agent` file-reference pattern. A complete briefing has these sections:
--- a/skills/spawn-agent/SKILL.md
+++ b/skills/spawn-agent/SKILL.md
@@ -22,6 +22,12 @@ Signals you want this skill:
 
 For a plain "open a named terminal for me" session, `twapp work --name foo` on its own is fine. Skip this skill.
 
+**Spawning a coordinator?** Use `twapp coordinator launch` and the
+[`agent-coordinator`](../agent-coordinator/SKILL.md) skill instead. The
+coordinator gets a dedicated command (canonical role metadata, bundled
+bootstrap, mailbox plumbing) that this generic worker-spawn path doesn't
+reproduce. This skill stays focused on workers.
+
 ## The file-reference pattern
 
 Inline prompts in `--run` are shell-fragile. A long prompt with curly quotes, `--flags`, backticks, or unicode arrows will be silently mangled by quoting before Claude ever sees it, and you'll spend ten minutes wondering why the agent is confused.
@@ -162,9 +168,13 @@ Include this in every briefing:
 - Within 2 minutes of spawn, write a hello message to
   `/tmp/agent-mailbox/inbox/<ISO-timestamp>-my-agent-hello.md`
   describing what you picked up and your planned first step.
+- Invoke `/loop` after hello; poll every 90-120s. Never finish a turn
+  asking the user a question; mailbox the coordinator if stuck.
 - Poll the inbox every 90s for follow-up messages addressed to your handle.
-- Invoke `/loop` after hello; poll every 90-120s. Never finish a turn asking the user a question — mailbox the coordinator if stuck.
 ```
+
+See [Headless operation](#headless-operation) below for the per-state
+defaults this rule encodes.
 
 Caller-side poll after spawn:
 
--- a/src-tauri/src/cli/coordinator.rs
+++ b/src-tauri/src/cli/coordinator.rs
@@ -272,7 +272,18 @@ pub fn resolve_mailbox(
     }
     if let Ok(v) = std::env::var("TWAPP_MAILBOX_DIR") {
         if !v.trim().is_empty() {
-            return Ok(PathBuf::from(v));
+            let inherited = PathBuf::from(&v);
+            // Don't refuse — other tools may set TWAPP_MAILBOX_DIR ahead of
+            // creating the directory — but warn loudly so a typo surfaces
+            // before the coordinator's first message lands in the void.
+            if !inherited.is_dir() {
+                eprintln!(
+                    "Warning: inherited TWAPP_MAILBOX_DIR={} is not an existing directory; \
+                     the spawned coordinator will get write errors on its first message.",
+                    inherited.display()
+                );
+            }
+            return Ok(inherited);
         }
     }
     let local = work_dir.join("mailbox");
@@ -423,6 +434,19 @@ pub fn find_session_dir(name: Option<&str>) -> Result<PathBuf, String> {
 mod tests {
     use super::*;
     use std::fs;
+    use std::sync::{Mutex, MutexGuard, OnceLock};
+
+    // Serialize every test that mutates TWAPP_MAILBOX_DIR. `cargo test` runs
+    // tests in parallel within a binary, and process-wide env is shared
+    // state — without this lock, `mailbox_creates_fallback_when_nothing_configured`
+    // and `mailbox_inherits_env_var` race and intermittently see each other's
+    // env mutation.
+    fn env_lock() -> MutexGuard<'static, ()> {
+        static LOCK: OnceLock<Mutex<()>> = OnceLock::new();
+        LOCK.get_or_init(|| Mutex::new(()))
+            .lock()
+            .unwrap_or_else(|p| p.into_inner())
+    }
 
     fn unique_tmp(prefix: &str) -> PathBuf {
         std::env::temp_dir().join(format!("{}-{}", prefix, uuid::Uuid::new_v4()))
@@ -608,6 +632,7 @@ mod tests {
 
     #[test]
     fn mailbox_creates_fallback_when_nothing_configured() {
+        let _guard = env_lock();
         let work_dir = unique_tmp("twapp-coord-mbx-fallback");
         fs::create_dir_all(&work_dir).unwrap();
         let prev = std::env::var("TWAPP_MAILBOX_DIR").ok();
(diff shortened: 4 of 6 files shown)


piekstra (commented):
Review — `coordinator launch | claim` is well-shaped; bundled template is right-sized; correctness is solid. A few non-blocking notes plus one heads-up about a coordinator directive that just landed in the mailbox.

## Policy scan

Clean on the extended regex. Bundled template uses generic role/handle vocabulary throughout.

## Heads-up: coordinator directive at 22:10:00Z

The coordinator sent a directive a few minutes ago to `twapp-coordinator-launch-cli` (and the parallel `twapp-session-role-metadata` / `twapp-model-selection` agents): **"new twapp subcommands must land alongside the skill docs that reference them, in the same PR."** This PR doesn't yet include changes to `skills/agent-coordinator/SKILL.md` documenting `twapp coordinator launch` / `twapp coordinator claim` (the skill currently teaches the manual `twapp work --name … --role …` pattern). Before self-merge:

- Add a section to `skills/agent-coordinator/SKILL.md` (probably near §1 "When to use" or §4 "Hello-within-2-min") explaining that `twapp coordinator launch` is the canonical entry point now, with the same example block your README has.
- Cross-link from the skill to the bundled `templates/coordinator-bootstrap.md` so future coordinators reading the skill know the bootstrap is what `launch` writes by default.

(Mailbox file: `20260420T221000Z-coordinator-to-twapp-coordinator-launch-cli-update-skills.md`. Worth a check that you've seen it — the skill update is the only thing standing between this PR's current state and ship-it.)

## Correctness

### `launch`

- **Existing-session guard.** `session_already_exists` checks for `.twapp-session.json` at the target dir before any state mutation. Returns exit 2 with a multi-line error pointing at `claim` and `stop` as the two recovery paths. Tested by `launch_refuses_existing_coordinator`. ✓
- **Briefing resolution.** `--briefing` path canonicalized + `is_file()` checked before spawn. Default path materializes the bundled template at `<work_dir>/.twapp-coordinator-bootstrap.md` and returns its absolute path. Both paths funnel through the same `Read <abs-path> and execute.` wrapper from the spawn-agent skill. ✓
- **Mailbox precedence (the briefing's 4-step rule).** Implemented faithfully:
  1. `--shared-dir` → use, create `<dir>/inbox`.
  2. `$TWAPP_MAILBOX_DIR` → inherit.
  3. `<work_dir>/mailbox/inbox/` exists → reuse.
  4. Otherwise create `<work_dir>/collab/mailbox/inbox/` and print to stderr.

  One small gap: step 2 does **no validation** of the inherited env var — a stale or typo'd `TWAPP_MAILBOX_DIR` will silently propagate to the spawned coordinator, who'll then get write errors on its first message. Worth at least a `Path::new(v).is_dir()` check + a single-line warning if missing. (Don't refuse — env vars set by other tools may not require pre-creation — just notice it.)

- **`build_run_command` shell escaping.** `TWAPP_MAILBOX_DIR='<escaped>' claude --dangerously-skip-permissions '<escaped>'` with `shell_escape_single` applied to both the mailbox path and the prompt. The bash escape (`'\\''`) inside the single-quoted strings is correct.
- **Step 1 also creates inbox/ but not archive/.** Agents need to write to `archive/` when archiving messages they've read; first archive call will create it on demand. Not blocking, but consider creating both for symmetry — it's one extra `create_dir_all` and prevents a subtle "where's my archive?" error.

### `claim`

- **Already-coordinator → no-op exit 0.** Idempotent. ✓
- **Non-coord without --force → exit 2 with explicit error**. Tested. ✓
- **Non-coord with --force → overwrites.** Tested.
- **JSON mutation via `serde_json::Value`** (not `SessionData`) preserves any unknown fields users might have manually added. Defensive — good. The `to_string_pretty` re-serialization keeps the file human-readable.
- **`find_session_dir`** scans `work_directory` for matching name; errors on 0-match or N>1-match. The N>1 error message is concrete ("Disambiguate by running `twapp coordinator claim` from the target directory."). ✓

### Forward-compat shim for `role`

- Same `#[serde(default, skip_serializing_if = "Option::is_none")]` shape as #43's plumbing. The two PRs' schemas are identical at the data model level — they will not conflict in the rebase regardless of merge order.
- **GUI fork sets `role: None`** (`gui/sessions.rs:1489`, `:1532`), matching the same divergence I flagged on #43 (CLI `cmd_resume --fork` clones role from parent; GUI fork doesn't). For this PR it's fine — the shim is only there to compile — but flagging for visibility:
  - If #44 merges first and #43 later adds GUI fork inheritance, the schema converges cleanly.
  - If #43 merges first WITH a GUI fork inheritance fix and #44 rebases, **make sure the rebase preserves #43's fork inheritance** — don't let `role: None` leak back into `fork_session`. A grep for `role: None` after rebase is the lazy check.
- **The `--role` validation in `cmd_work` and `create_session_core`** rejects empty/whitespace strings (matching what #43 does in `validate_role`). Same logic, different home — the duplication is fine for the merge-order independence the PR body promised.

### Bundled template

- 63 lines. Right-sized — points coordinators at the skill (§1), the messaging design (§4), and the mailbox protocol without re-stating them.
- §3 cadence guidance: "Do not poll faster than ~90s". Matches the agent-coordinator skill's own §3 cadence. ✓
- §3 "Fetch blockers first" via `twapp msg fetch --priority blocker` reads the messaging-design's PR-4 priority lane — correctly aspirational for now (PR-1 ships flat priorities; PR-4 adds the urgent dir). The `twapp msg fetch` command itself exists today after #41, so the example works modulo the `--priority blocker` flag.
- `include_str!` embeds at compile time → updates require a rebuild + re-launch of the coordinator. That's right; an on-disk path would be a footgun (someone editing the bundled template would change behavior for every new coordinator).

## Test coverage

7 tests, all genuinely useful:

| Test | Verifies |
|---|---|
| `launch_writes_role_coordinator` | Schema contract + `COORDINATOR_ROLE` constant matches literal `"coordinator"` |
| `launch_without_briefing_uses_default_template` | Bundled template materialized + absolute path + content match |
| `launch_refuses_existing_coordinator` | Guardrail predicate |
| `claim_flips_role_in_place` | Happy path |
| `claim_refuses_overwriting_role_without_force` | Both refuse and force paths in one test |
| `mailbox_prefers_shared_dir_arg` | `--shared-dir` precedence |
| `mailbox_creates_fallback_when_nothing_configured` | Final fallback path |

**Missing (worth adding):**

- `mailbox_inherits_TWAPP_MAILBOX_DIR_env_var` — the second-precedence rung is currently untested.
- `mailbox_reuses_existing_local_mailbox` — the third-precedence rung (work_dir/mailbox with inbox/ already populated).
- `claim_with_name_disambiguates_correctly` — covers the multi-match error path in `find_session_dir`.

**Concurrency note:** `mailbox_creates_fallback_when_nothing_configured` mutates `TWAPP_MAILBOX_DIR` env var directly without serialization. Same concern as #42's `HOME_LOCK` pattern — `cargo test` runs tests in parallel within a binary, so if any future test in this file also touches that env var (the missing inherit-test above will), they'll race. Wrap both tests with a `static MAILBOX_ENV_LOCK: Mutex<()> = Mutex::new(());` once you add the inherit test.

## Non-blocking nits

- **`launch`'s `eprintln!("Using mailbox: {}", mailbox.display())`**: helpful, but consider gating behind a `--verbose` flag or moving to `log::info!` so scripted spawns don't get unexpected stderr noise. Not blocking.
- **`sanitize_name`'s underscore-to-dash replacement** is applied after `split_whitespace().join("-")`. Means `--name "my_worker session"` → `my-worker-session`. Surprising — the underscore-to-dash transformation isn't documented anywhere. Either keep underscores as-is or document the transformation in the `--name` help text.
- **`templates/coordinator-bootstrap.md` §2 step 2** asks the coordinator to maintain `<shared-dir>/handle.txt` manually. The agent-coordinator skill §12 also references `handle.txt`. Worth a future PR-scoped step to add `twapp coordinator register-handle` so this isn't a manual file mutation forever — but out of scope for this PR.

## Summary

Coordinator launch + claim, bundled template, mailbox precedence — all match the briefing and the design doc. Forward-compat shim for `role` is structurally identical to #43's; merge order is flexible.

**Two things before flipping to ship-it:**
1. Address the coordinator directive at 22:10:00Z by including skill-doc updates for `twapp coordinator launch | claim` in this PR.
2. Either validate `$TWAPP_MAILBOX_DIR` on inherit (warn-on-missing), or document that step 2 of mailbox precedence is trust-the-env-var.

Everything else is non-blocking polish. Happy to flip to ship-it as soon as the skill-doc updates land — the code itself is in good shape.

piekstra (commented):
**Re-review at tip `2aa7981c` — code itself is excellent, but the DIRTY status hides a substantial post-rebase rework. Code-side ship-it conditional on the rebase being clean.**

## Policy scan

Clean on the extended regex.

## What you addressed

All four review items from my prior comment are nailed:

1. **Skill docs land with the subcommand** (coordinator directive 22:10:00Z): agent-coordinator §1b "Bootstrapping a coordinator" + spawn-agent §"When to use" cross-reference. ✓ The cross-reference is the right shape — it points readers at the canonical command without duplicating the agent-coordinator skill's content.
2. **Headless operation codified** (coordinator directive 22:51:00Z): agent-coordinator §2.5 with the full rule + briefing template one-liner + spawn-agent Protocol example block + a load-bearing-rationale paragraph. ✓
3. **Stale TWAPP_MAILBOX_DIR warn**: non-fatal `eprintln!` warning when the inherited path isn't a directory. Right level — keeps the "trust the env var" behavior while surfacing typos. ✓
4. **Test coverage**: `mailbox_inherits_env_var_when_set`, `mailbox_reuses_existing_local_mailbox`, `find_session_dir_errors_on_empty_cwd`, plus `env_lock()` mutex pattern matching #46/#48's convention. ✓

The `find_session_dir_errors_on_empty_cwd` test mutates `std::env::current_dir` mid-test — this could in principle race with another cwd-touching test. Probably fine since no other test in this file touches cwd, but worth keeping under the same `env_lock()` if any future test needs cwd manipulation.

## ⚠️ Real blocker: rebase is non-trivial against merged main

`mergeStateStatus: DIRTY`. The conflicts aren't just textual — they're structural overlaps with already-merged work:

### `src-tauri/src/cli/session.rs` — `role` field

Your branch's diff shows adding `role: Option<String>` to `SessionData` (8 lines additive). **Main already has this** (from merged #43, which also added `provenance: Option<String>` and the `--role`/`--spawned`/`--provenance` flag set on `twapp work`). After rebase, **drop the entire shim**:
- Remove the duplicate `role` field declaration in `SessionData`.
- Remove the `--role` flag declaration on `Commands::Work` (already in main).
- Remove the duplicated `role` parameter threading in `cmd_work` and the inline `--role cannot be empty` validation (#43's `validate_role` is already in main).

### `src-tauri/src/cli/mod.rs::create_session_core` call site

Your `cmd_launch` calls `create_session_core` with **8 args**:

```rust
super::create_session_core(
    None, Some(session_name.clone()), Some(run_command),
    false, None, None, false,
    Some(COORDINATOR_ROLE.to_string()),
)
```

**Main now expects 9 args** — #43 added `provenance: Option<String>` after `role`. After rebase, this call needs:

```rust
super::create_session_core(
    None, Some(session_name.clone()), Some(run_command),
    false, None, None, false,
    Some(COORDINATOR_ROLE.to_string()),
    Some("spawned".to_string()),  // coordinator launches itself programmatically
)
```

`"spawned"` is the right provenance for a coordinator spawned via `twapp coordinator launch` — a human typing `twapp coordinator launch` is initiating, but the resulting session is one that `twapp` (not Claude) bootstraps with a briefing file. Could argue either way; I'd default to `"spawned"` to match the `--from-file` heuristic from #43.

### `launch_writes_role_coordinator` test

The test builds a `SessionData` literal and lists fields up to `role: Some(COORDINATOR_ROLE.to_string())`. Add `provenance: Some("spawned".to_string())` (or `None`) at the end after rebase.

### `skills/agent-coordinator/SKILL.md` — §2.5 "Headless operation"

**Main already has this** (from merged #47). Your §2.5 wording is similar but not identical. The §2 Protocol template's `/loop` one-liner is also in main from #47. After rebase you have three options:
- Drop your §2.5 and §2 Protocol additions; let #47's already-merged content stand.
- Merge content: keep #47's section but add your "load-bearing rationale" paragraph if it adds value not in #47's version.
- Replace #47's content with yours if it's clearly superior (it's a wash IMO; #47's framing of "coordination failure not worker failure" is a nice angle yours doesn't have).

The cleanest path is: **drop your duplicate §2.5 and Protocol-template additions; keep your new §1b "Bootstrapping a coordinator" + spawn-agent cross-reference + the `Why /loop is load-bearing` paragraph in spawn-agent** (which doesn't overlap with #47).

### `skills/spawn-agent/SKILL.md`

Main has #47's "Headless operation" top-level section. Your branch adds a `/loop` line to the Protocol example block + a "Why /loop is load-bearing" paragraph. After rebase: keep both your additions (they don't duplicate #47's section, they're inline reinforcements).

#43 also touched spawn-agent (Examples A and B, plus the "Coming soon" removal). Should auto-merge cleanly with your additions since the regions are different.

### `templates/coordinator-bootstrap.md`

New file. No conflict.

## Other notes (non-blocking)

- **`create_session_core` provenance ambiguity** (above): could argue `"user"` since a human runs `twapp coordinator launch`. I'd defer to the agent-aware-ui design doc's framing — `"spawned"` if the launching subject is "another twapp session," `"user"` if a person typed it. A coordinator is *typed by a human* but operates programmatically afterward. Pick one and document it in the `cmd_launch` docstring so future maintainers see the intent.
- **Coordinator skill §1b** is great — explicit framing that `twapp coordinator launch` is canonical and the manual `twapp work --name coordinator …` invocation is the fallback. Right hierarchy.
- **Bundled `templates/coordinator-bootstrap.md`** points at this skill, this skill points at the bundled template — bidirectional links keep both fresh.

## Tests not in scope (follow-ups already noted in your message)

- `--verbose`-gated stderr noise: noted, defer.
- Underscore-to-dash sanitize doc: noted, defer.
- `twapp coordinator register-handle` automation for `handle.txt`: agreed, separate scope.

## Summary

**Code substance: ship-it.** All four review items addressed cleanly, plus both coordinator directives. The real blocker is now mechanical: the rebase against current main needs to:

1. Drop the `role` field shim from `SessionData` (already in main from #43).
2. Drop the `--role` flag definition from `Commands::Work` (already in main from #43).
3. Drop the inline `validate_role` calls in `cmd_work` (already in main from #43's `validate_role` + `resolve_provenance`).
4. Add a 9th `provenance: Some("spawned".to_string())` arg to the `create_session_core` call in `cmd_launch`.
5. Add the `provenance` field to the `SessionData` literal in the `launch_writes_role_coordinator` test.
6. Drop §2.5 "Headless operation" from `agent-coordinator/SKILL.md` (#47 already added equivalent content; your §2 Protocol template line is also redundant with #47).
7. Keep your new `agent-coordinator` §1b + spawn-agent §"When to use" cross-reference + spawn-agent Protocol-block `/loop` line + "Why /loop is load-bearing" paragraph — those don't conflict.

Once the rebase lands and CI is green, this is unconditional ship-it. The substance is right, the scope is right, the doc updates honor both coordinator directives. Just needs the mechanical rebase to align with what's already in main.

piekstra (commented):
**Ship it.** Re-reviewed at tip `c28b61f2` — all 7 rebase items from my prior review landed cleanly.

## Policy scan

Clean.

## CI

`PR Check` SUCCESS at 02:20:11Z. ✓

## What the rebase did

Verified each of the 7 post-rebase items from my prior review against the current diff:

1. ✅ **`role` field shim dropped** — `src-tauri/src/cli/session.rs` is no longer modified. #43's merged `role` + `provenance` fields on `SessionData` are used directly.
2. ✅ **`--role` flag definition dropped** — `src-tauri/src/cli/mod.rs` no longer declares the flag on `Commands::Work`. #43's merged version handles it.
3. ✅ **Inline `--role` validation dropped** — `cmd_work` no longer has the `if r.trim().is_empty()` branch; #43's `validate_role` in main does that.
4. ✅ **`provenance: Some("spawned")` 9th arg added** to `create_session_core` call in `cmd_launch`. The full call is now `(None, Some(session_name), Some(run_command), None, false, None, None, false, Some(COORDINATOR_ROLE.to_string()), Some("spawned".to_string()))`. Matches merged main's signature. `"spawned"` is the right provenance — the coordinator runs a briefing file, so it's programmatic even though a human typed the `launch` command.
5. ✅ **`launch_writes_role_coordinator` test's `SessionData` literal** compiles against merged main (CI green confirms).
6. ✅ **Duplicate §2.5 "Headless operation" dropped** — `skills/agent-coordinator/SKILL.md` diff is now +36/-0 (was +59/-1), matching the "keep only §1b + cross-refs + load-bearing paragraph" plan. `agent-coordinator` already carries #47's headless-op content from main.
7. ✅ **Kept**: new §1b "Bootstrapping a coordinator" (agent-coordinator), spawn-agent §"When to use" cross-reference to agent-coordinator, spawn-agent Protocol `/loop` line + "Why /loop is load-bearing" paragraph. None overlap with #47's content; all genuinely additive.

## Net delta against pre-rebase branch

| File | Pre-rebase | Post-rebase | Delta |
|---|---|---|---|
| README.md | +50 | +50 | 0 (docs unchanged) |
| skills/agent-coordinator/SKILL.md | +59/-1 | +36/0 | §2.5 duplication dropped |
| skills/spawn-agent/SKILL.md | +15 | +11/-1 | simplified |
| src-tauri/src/cli/coordinator.rs | +710 | +719 | +9 (provenance arg + call-site tidy) |
| src-tauri/src/cli/mod.rs | +37/-2 | +19/0 | `--role` flag removed |
| src-tauri/src/cli/session.rs | +8 | 0 | `role` field shim removed |
| templates/coordinator-bootstrap.md | +63 | +63 | 0 |

Clean rebase.

## One observation (non-blocking)

`mergeStateStatus: UNKNOWN` at the moment I checked — likely still computing. The merge-state should resolve to `CLEAN` or `BEHIND` within a minute; if it comes back `DIRTY` there's another conflict I missed. CI being green means no code conflicts; README + skills should merge textually without issue given the clean rebase above.

## Summary

Coordinator launch/claim feature, bundled bootstrap template, mailbox precedence with stale-env-var warning, comprehensive tests including the cross-imports from #43's `role`/`provenance` plumbing. Skill-doc updates honor coordinator's directives 22:10Z + 22:51Z without duplicating #47's already-merged headless-op content.

**Ship it.** Coordinator can self-merge; I'd suggest a final `gh pr view 44 --json mergeStateStatus` check before merge just to confirm UNKNOWN → CLEAN transition. The substance is solid and CI is green.

piekstra (commented):
**Still ship-it at tip `62c3a5a7`** — this is a reactive rebase after #45 merged. No code changes; README changes are exactly what should land.

## What changed since my last ship-it (`c28b61f2` → `62c3a5a7`)

- **README**: the `### Bootstrapping a coordinator` subsection under #45's co-lab umbrella used to be a "Coming in #44" forward-reference placeholder. This rebase **replaces the placeholder with the actual coordinator launch/claim documentation** — which is precisely the cross-PR plan that #45 and #44 were coordinating on. Clean completion of the umbrella story.
- **CLI Reference table**: two new rows (`twapp coordinator launch` + `twapp coordinator claim`) spliced in after #48's `twapp msg claim/release` rows.
- **Everything else unchanged**: `coordinator.rs +719`, `mod.rs +19`, `spawn-agent/SKILL.md +11/-1`, `agent-coordinator/SKILL.md +36`, `templates/coordinator-bootstrap.md +63` — all identical to the prior tip I shipped.

## Net effect

The previously-noted 7 rebase items stay addressed (role field shim dropped, `--role` flag dropped, provenance threaded, duplicate headless section dropped, etc.) AND the README placeholder from #45 is now filled in with real content. This is the mergeable state.

## Status

- Policy scan: clean.
- CI: `IN_PROGRESS` at 02:26:37Z — expected to go green given no code changes from the prior tip I shipped.
- `mergeStateStatus: UNSTABLE` only because CI is running.

**Ship it.** Once CI completes (green), coordinator can self-merge. This closes out the coordinator-launch story cleanly with #45's co-lab umbrella.

PR #46, opened 2026-04-20
feat: model selection (--model) + models list/refresh subcommands

piekstra:
## Summary

- `twapp work --model <name>` pass-through to the spawned provider CLI. Claude gets `--model <name>`; codex gets `-c model='<name>'`. When unset, twapp does not pass `--model` and the provider's own default wins. No validation — the provider CLI rejects unknown names.
- `twapp models list [--provider <p>] [--format json]` prints a three-column `NAME / TIER / DESCRIPTION` table (or JSON). Reads from `~/.config/twapp/models.<provider>.json` if cached, otherwise falls back to a bundled default committed at `data/models.claude.default.json` so the command works on a fresh install with no network call.
- `twapp models refresh [--provider <p>]` (claude only for now) hits `https://api.anthropic.com/v1/models` with `x-api-key` + `anthropic-version: 2023-06-01`, parses `data[]`, and writes the cache atomically (temp + rename). Requires `ANTHROPIC_API_KEY`; errors cleanly if unset. Codex refresh errors with a pointer to the cache path for manual editing until the upstream CLI exposes a listing endpoint.
- Skill docs (`skills/spawn-agent/SKILL.md`, `skills/agent-coordinator/SKILL.md`) and README updated alongside the feature per coordinator request — future agents can discover the verb without a lag.

## Design notes

- Model list is loaded from JSON (both bundled default and cache), not hardcoded in Rust, so editors don't need to rebuild to change the list.
- Command-building extracted into `build_claude_run_command` / `build_codex_run_command` helpers in `cli/session.rs` so the pass-through can be unit-tested directly without mocking the spawn side.
- Explicitly out of scope: `--model` on `twapp resume` (resume keeps whatever model the original spawn picked) and on `twapp coordinator launch` (subcommand not yet landed in main — will fold in via the `twapp-coordinator-launch-cli` lane's PR or a follow-up).

## Test plan

- [x] `cargo test --lib` — 50 passing, incl. 6 new `cli::models::tests::*` and 5 new `cli::session::command_build_tests::*`
  - `model_flag_threads_through_to_claude_invocation` — asserts `--model 'claude-sonnet-4-6'` appears in the generated claude command
  - `claude_invocation_omits_model_when_unset` — asserts no `--model` when flag is absent
  - `model_flag_threads_through_fork_resume` — asserts `--model` is positioned correctly on fork-resume
  - `model_flag_threads_through_to_codex_invocation` — asserts `-c model='o3'` appears in the generated codex command
  - `codex_invocation_omits_model_when_unset`
  - `bundled_default_parses` — sanity check on the shipped JSON
  - `models_list_loads_bundled_default_when_cache_missing`
  - `models_list_prefers_cache_over_bundled_when_both_exist`
  - `models_refresh_parses_anthropic_response` — parses a canned fixture, asserts tier mapping
  - `models_refresh_errors_without_api_key`
  - `tier_for_name_matches_family`
- [x] `cargo clippy --lib` — no new warnings in touched files
- [x] Smoke: `twapp models list`, `twapp models list --format json`, `twapp models refresh` (errors without key), `twapp models refresh --provider codex` (errors with manual-edit pointer), `twapp work --help` (shows `--model`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -479,12 +479,83 @@ Without these checks, the spawned terminal would appear to launch fine and
 then silently fail inside the new window — a painful debug loop when
 automating spawns.
 
+## Model selection
+
+`twapp work --model <name>` pass-through sets the model on the spawned
+provider CLI. When unset, twapp does **not** pass `--model` and the
+provider's own default applies (e.g. the Claude CLI's global
+`ANTHROPIC_MODEL` or user config).
+
+```bash
+# Pin a spawned worker to a specific Claude model.
+twapp work --name plumbing-worker \
+  --from-file /path/to/briefing.md \
+  --model claude-haiku-4-5-20251001
+
+# Use the tier alias for the latest model of that tier.
+twapp work --name design-worker --from-file /path/to/briefing.md --model opus
+```
+
+twapp does **not** validate the model name — pass-through means the
+provider CLI rejects unknown names at spawn time. For claude, the value
+is forwarded as `--model <name>`; for codex, as `-c model='<name>'`
+(a TOML config override). The `twapp resume` command intentionally does
+not accept `--model` — a resumed session keeps whatever model the
+original spawn picked.
+
+### Discovering available models
+
+```bash
+# Three-column table: NAME / TIER / DESCRIPTION.
+twapp models list
+
+# Different provider (defaults to claude).
+twapp models list --provider claude
+
+# Machine-readable output for scripts.
+twapp models list --format json
+
+# Refresh the cache from the provider's models endpoint.
+# For claude, requires ANTHROPIC_API_KEY in the environment.
+ANTHROPIC_API_KEY=sk-ant-... twapp models refresh
+```
+
+`twapp models list` reads a cache at
+`~/.config/twapp/models.<provider>.json` if present, or falls back to a
+bundled default list shipped with the binary.
+
+> **Caveat:** the bundled default is a snapshot of the Claude model
+> family at build time. `twapp models refresh` is the authoritative
+> source — run it whenever you want a current view of available models.
+> The cache always takes precedence over the bundled default.
+
+`twapp models refresh` currently supports `--provider claude` (calls
+`https://api.anthropic.com/v1/models` with `x-api-key` and
+`anthropic-version: 2023-06-01`). For codex, edit the cache file by
+hand; a refresh verb will land when the upstream CLI exposes a listing
+endpoint.
+
+### Picking a tier when spawning workers
+
+- **haiku** — plumbing, doc edits, simple refactors, CLI scaffolding.
+- **sonnet** — default for most implementation work; good balance of
+  capability and cost.
+- **opus** — design audits, cross-cutting synthesis, high-stakes
+  correctness work where the model's reasoning is load-bearing.
+
+Match the model to the scope cost. Sending a one-line dependency bump
+to opus burns budget; sending a complex architecture audit to haiku
+burns iteration.
+
 ## CLI Reference
 
 | Command | Description |
 |---------|-------------|
 | `twapp work <ticket\|--name>` | Start a new work session using the configured provider |
 | `twapp work --from-file <path>` | Spawn a session whose prompt is `Read <path> and execute.` (safer than `--run` for long prompts) |
+| `twapp work --model <name>` | Pass-through model selection; forwarded to the provider CLI (claude: `--model`, codex: `-c model='…'`) |
+| `twapp models list [--provider <p>] [--format json]` | Show known models for the provider (cache if present, else bundled default) |
+| `twapp models refresh [--provider <p>]` | Re-populate the provider cache from the models endpoint (claude: `ANTHROPIC_API_KEY` required) |
 | `twapp stop --name <name> [--force]` | Gracefully stop a running session (SIGTERM, optional SIGKILL escalation) |
 | `twapp resume [--fork]` | Resume or fork the current session using the configured provider |
 | `twapp sessions` | List all sessions with activity timestamps |
--- a/data/models.claude.default.json
+++ b/data/models.claude.default.json
@@ -1,0 +1,32 @@
+[
+  {
+    "name": "claude-opus-4-7",
+    "tier": "opus",
+    "description": "Most capable Claude model; use for design audits and complex synthesis."
+  },
+  {
+    "name": "claude-sonnet-4-6",
+    "tier": "sonnet",
+    "description": "Strong general-purpose Claude model; default for most implementation work."
+  },
+  {
+    "name": "claude-haiku-4-5-20251001",
+    "tier": "haiku",
+    "description": "Fast and inexpensive Claude model; use for plumbing and simple edits."
+  },
+  {
+    "name": "opus",
+    "tier": "opus",
+    "description": "Alias: latest opus-tier Claude model."
+  },
+  {
+    "name": "sonnet",
+    "tier": "sonnet",
+    "description": "Alias: latest sonnet-tier Claude model."
+  },
+  {
+    "name": "haiku",
+    "tier": "haiku",
+    "description": "Alias: latest haiku-tier Claude model."
+  }
+]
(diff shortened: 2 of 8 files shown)


piekstra (commented):
Review — feature scope is right, command-build extraction is clean, skill-doc updates land alongside per coordinator's directive. One real concern (env-var test in `unsafe` form) plus a few non-blocking notes.

## Policy scan

Clean on the extended regex. Model identifiers are real Claude model names (`claude-opus-4-7`, `claude-sonnet-4-6`, `claude-haiku-4-5-20251001`) + tier aliases (`opus`/`sonnet`/`haiku`).

## Coordinator directive compliance ✓

The 22:10:00Z directive said "new twapp verbs/flags must land alongside the skill doc updates that reference them, in the same PR." This PR delivers exactly that:

- `skills/spawn-agent/SKILL.md` +43 lines: "Model selection" subsection with command examples, sample `twapp models list` output, when-to-pick-which-tier guidance.
- `skills/agent-coordinator/SKILL.md` +12 lines: "Model selection when spawning workers" paragraph in §2 that defers detail to spawn-agent (right cross-link discipline — single source of truth).
- README +71 lines: top-level "Model selection" section + CLI Reference entries.

Discovery is now possible from any of the three surfaces.

## Correctness

### `build_claude_run_command` / `build_codex_run_command` extraction

Right call. Pulling these out of `create_session_core` makes flag threading directly unit-testable. The five `command_build_tests` cover:
- claude with model set → `--model 'X'` appears
- claude without model → no `--model` token
- claude fork-resume with model → cd-prefix + correct flag order
- codex with model → `-c model='X'` appears
- codex without model → no `-c model=` token

The shell-escape on the model name (`shell_escape_single`) is defensive — current model names don't need quoting, but a future name with shell metacharacters wouldn't break the spawn. ✓

`--model` is positioned right after `claude` / `codex` (before `--resume` / `-C` / etc.). Order doesn't matter to either CLI, so this is fine.

### Bundled default + cache + refresh flow

- **`include_str!`** for the bundled default — compile-time embed; users can't accidentally edit the bundled list and confuse themselves. ✓
- **`load_models_from`** prefers cache over bundled. Tested both ways.
- **`write_cache_atomic`** uses temp + rename. Won't leave a half-written cache file on partial failure. ✓
- **`tier_for_name`** is substring-based — `opus`/`sonnet`/`haiku` in lowercase, otherwise `"other"`. Heuristic but correct for current model naming. Future Claude models presumably keep the family-name convention.
- **`parse_anthropic_response`** only deserializes the success shape. Anthropic API errors (JSON like `{"type":"error",...}`) won't deserialize cleanly, but `curl -f` exits non-zero on HTTP errors, so the error path triggers at the curl layer first. Both layers are correct in isolation.
- **`refresh_claude`** shells out to `curl` rather than using a Rust HTTP client. Probably reflects an existing project convention (no reqwest dep visible in the diff). Acceptable; macOS and most Linux distros ship curl.
- **`-sS -f`** is the right flag combo: `-s` silent, `-S` show errors, `-f` fail on HTTP errors without printing the body. The error message "curl exited with status N" doesn't include the API's error body, which is mildly opaque but not blocking.

### `cmd_list` table formatter

Computes `name_w` and `tier_w` as the max-of-column-or-4. ✓ Sensible width math. Source line goes to stderr, table to stdout — pipe-friendly. ✓ Empty cache prints `"no models cached for X (source: bundled)"` to stdout (not stderr) — minor inconsistency with the stderr source line below it; not blocking.

### `--format` arg

Plain `Option<String>` with `Some("json")` matched. Unknown values exit 2 with a clear error. Could've been `clap::ValueEnum` to push validation to clap, but the current shape works.

## One real concern: `models_refresh_errors_without_api_key` test compile

```rust
let prior = std::env::var("ANTHROPIC_API_KEY").ok();
std::env::remove_var("ANTHROPIC_API_KEY");
let code = refresh_claude();
if let Some(v) = prior {
    std::env::set_var("ANTHROPIC_API_KEY", v);
}
```

PR #42 (already merged) wraps `std::env::set_var` and `std::env::remove_var` in `unsafe { ... }` blocks, and inside an `impl Drop` to ensure restoration even on panic. That suggests this codebase compiles on Rust 1.82+, where `set_var` and `remove_var` are marked unsafe due to thread-safety. **This test should fail to compile** unless the project's MSRV is older.

CI is currently `IN_PROGRESS`; if it goes red, this is the cause. Fixes:

1. Wrap in `unsafe { ... }` to mirror #42's pattern:
   ```rust
   unsafe { std::env::remove_var("ANTHROPIC_API_KEY"); }
   // ...
   unsafe { std::env::set_var("ANTHROPIC_API_KEY", v); }
   ```

2. Or, since the test panics-could leave HOME unset for other parallel tests, port #42's `HomeGuard` pattern:
   ```rust
   struct ApiKeyGuard { prior: Option<String> }
   impl ApiKeyGuard { fn clear() -> Self { /* save prior, unsafe remove */ } }
   impl Drop for ApiKeyGuard { fn drop(&mut self) { /* unsafe restore */ } }
   ```
   This also guards against test-panic-leaves-state-unset.

Worth checking CI before merge regardless.

## Test infrastructure note (non-blocking)

`models_refresh_errors_without_api_key` mutates `ANTHROPIC_API_KEY` env var directly. Same race risk I flagged on #42 — `cargo test` parallelizes within a binary, so any future test in this file (or in `cli::session::command_build_tests` if they ever read this env var) could race. Once the unsafe wrapping is in, also consider adding a `static API_KEY_LOCK: Mutex<()> = Mutex::new(());` and claiming it at the top of any test that touches the env var.

## Coordination with parallel PRs

Per the author's mailbox note:

- **#43 (`twapp-session-role-metadata`)** — both add a flag to `Commands::Work`. The placement in this PR (`model: Option<String>` after `from_file`, before `github`) is fine; #43 puts `role`/`spawned`/`provenance` after `chrome`. Whichever lands second rebases trivially — different argument positions, no conflict. ✓
- **#44 (`twapp-coordinator-launch-cli`)** — coordinator note says `--model` on `twapp coordinator launch` is deferred to whichever PR lands second. Confirmed in #44's body. ✓
- **`create_session_core` signature** gains `model: Option<String>` as the 4th positional. GUI caller in `gui/sessions.rs::create_and_launch_session` is updated to pass `None`. #44's `cmd_launch` calls `create_session_core` with `Some(COORDINATOR_ROLE.to_string())` for the role parameter — but that's the role parameter, not model. Whichever lands second will need to thread an extra `None` (or a real model) through that call. Mechanical rebase.

## Tests not in scope but worth a follow-up

- `cmd_list` happy-path tests (with a seeded cache) and `--format json` happy path.
- `cmd_list` with `Some("xml")` → exit 2.
- `cmd_refresh` with a mocked curl response. Hard without dependency injection — fine to defer.
- An end-to-end test that asserts `--model` survives the round-trip from `cmd_work` flag → `create_session_core` → `build_claude_run_command`. Currently the unit boundary is well-tested but the integration is implicit.

## Summary

Clean feature, right cross-doc discipline, sensible test coverage. Two things to address before flipping to ship-it:

1. **Wrap the env-var mutations in `models_refresh_errors_without_api_key` in `unsafe { ... }`** — match #42's pattern. CI should catch this if MSRV requires it.
2. **(Optional)** Guard the env-var test with a `Mutex` so future env-var tests don't race.

Everything else is non-blocking polish. Ship it as soon as CI is green and the unsafe wrap is in.

piekstra (commented):
**Update: CI green — flipping to ship-it.**

`PR Check` on `0c9c681d` came back SUCCESS, so my speculation about Rust 1.82+ requiring `unsafe` wrapping for `std::env::set_var` / `remove_var` doesn't apply on this codebase's MSRV. Disregard that concern as a blocker.

**Future-proofing note (non-blocking):** if someone later bumps the project MSRV to 1.82+, `models_refresh_errors_without_api_key` will start failing to compile (and any new env-var-mutating tests will too — see `cli/session_attribution.rs` in #42 for the pattern that already uses `unsafe { }`). Worth either:
- Pre-emptively wrapping the env mutations in `unsafe { }` now (compiles on both old + new Rust), or
- Leaving as-is and dealing with it if/when MSRV bumps. CI today confirms it compiles cleanly.

Either is fine.

**Everything else from my prior review still stands:**
- Wire format / cache flow / atomic write — solid.
- `build_claude_run_command` / `build_codex_run_command` extraction unit-testable, 5 tests pin the threading.
- Skill-doc updates land alongside per coordinator's directive (spawn-agent +43, agent-coordinator +12, README +71). Single source of truth (spawn-agent), agent-coordinator defers via cross-link. Right discipline.
- No conflict with #43 / #44 — different argument positions; mechanical rebase whichever lands second.

**Ship it.**

PR #47, opened 2026-04-20
docs(skills): mandate headless operation via /loop

piekstra:
## Summary

- Add a **Headless operation** section to `skills/spawn-agent/SKILL.md` with the shared contract phrasing and per-state defaults (waiting-for-review, concern-not-Ship-it, CLI-rejected, work-complete). Thread the `/loop` mandate into the briefing-template Protocol block so every spawn-agent-produced briefing inherits it.
- Add a matching **Headless operation — worker contract** subsection under §2 *Briefing structure* in `skills/agent-coordinator/SKILL.md`, frame a worker-asks-the-user-a-question turn as a coordination failure (not a worker failure), and insert the same one-liner into the canonical briefing template's Protocol block so coordinator-written briefings inherit the rule.
- Both skills now carry the same canonical sentence in matching phrasing, so readers of either file see the same contract.

### Why

Earlier today an implementer finished a turn and asked the user "Do you want me to poll?" — blocking for ~9 minutes. The coordinator broadcast a runtime correction to running agents, but the merged skills themselves still lacked the rule, so a newly-spawned agent would repeat the mistake. This closes that gap at the source.

### Out of scope

- No `src/` or `src-tauri/` touches; docs-only.
- No rewrites of other skill sections.
- No changes to `docs/designs/`.

## Test plan

- [x] `skills/spawn-agent/SKILL.md` has a **Headless operation** top-level section.
- [x] `skills/agent-coordinator/SKILL.md` §2 has **Headless operation — worker contract** subsection.
- [x] Briefing-template Protocol blocks in both skills include the `/loop` one-liner.
- [x] Canonical sentence is identical across both skills ("Every worker invokes `/loop` after its hello and operates autonomously…").
- [x] No private-project references (grep clean).
- [ ] Reviewer confirms markdown renders cleanly in the GitHub preview.

## Coordinate with

- #44 (`twapp-coordinator-launch-cli`) — also touches skills. If #44 lands first I will rebase on updated skills and refine; if this lands first, #44 rebases.
- #45 (`twapp-readme-colab`) — README only; no conflict expected.

PR #48, opened 2026-04-20
feat: lane-claim coordination primitive (twapp msg claim/release)

piekstra:
## Summary

- Adds `twapp msg claim <lane-id>` / `release <lane-id>` / `claim --list` — a filesystem-atomic lane-claim primitive for N workers pulling from a shared queue (reviewers on PRs, auditors on backlogs, implementers on task lists).
- Atomic resolver is POSIX `mkdir` of `<mailbox>/claims/<lane-id>/`; `owner.json` is written tmp+rename inside. Release writes a sibling `released.json`, leaves the dir in place as audit trail; next claim archives it to `<lane-id>.released-<ts>/`.
- Every claim / reclaim / release emits a `to: [all]` shadow broadcast into the mailbox inbox so the event is visible in the normal message flow — the "I got this, do you ack?" handshake is visible to humans + other agents at the same time as the on-disk lock.
- Stale reclaim (`--stale-seconds`, default 600) overwrites `owner.json` with `reclaimed_from: <previous-owner>` and emits a reclaim broadcast; this is the only non-mkdir-atomic path and is coarse-grained by design.
- Skills + docs: `agent-coordinator` gets §8.5 "N-worker coordination via lane claims" with a reviewer-race example; `spawn-agent` cross-refs it; new `docs/designs/worker-coordination.md` covers atomic-mkdir rationale, stale semantics, and out-of-scope list.

## Test plan

- [x] `cargo test --lib` — 68 passed (18 new in `cli::msg_claim::tests`).
- [x] Concurrent claim: 8-thread race with `std::sync::Barrier`; exactly one `Claimed` outcome, 7 `AlreadyClaimed`.
- [x] Release refuses non-owner + unclaimed lane.
- [x] Re-claim after release: previous dir is archived under `<lane>.released-<prev-claimed-at>`.
- [x] Stale-reclaim by seeded past `claimed_at` (no sleep — deterministic).
- [x] List excludes released + stale + non-matching prefix; emits valid JSON.
- [x] End-to-end CLI smoke: fresh claim → contest → list → release → re-claim → list --format json → 3 broadcasts in inbox + archive dir on disk.

## What's NOT in scope

Explicit in `docs/designs/worker-coordination.md` §3: distributed locking across hosts, cryptographic ownership proof, priority/preemption of claims, byzantine workers, GitHub review-assignment integration, periodic archive rotation (relies on the existing mailbox janitor).

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,6 +417,42 @@ get synthetic defaults (routine priority, no thread, id derived from the
 filename). This keeps day-one upgrades unbreaking; directory split and
 layout migration live in later PRs.
 
+### Lane claims — coordinating N workers on a shared queue
+
+When two or more sessions pull from the same list — reviewers against a
+PR queue, auditors against a backlog, implementers pulling from a
+prioritized task file — `twapp msg claim` / `release` adds an atomic
+"I got this" step before the work, so simultaneous workers don't
+double-up on the same item. The primitive is a POSIX-atomic `mkdir`
+into `<mailbox>/claims/<lane-id>/`, an `owner.json` written
+(tmp+rename) inside, and a `to: [all]` broadcast into the inbox so the
+event shows up in the normal message flow.
+
+```bash
+# Before each item: atomic claim. Exit 0 → proceed; exit 1 → skip.
+if twapp msg claim PR-91 --note "reviewing"; then
+  gh pr view 91
+  # ...post review...
+  twapp msg release PR-91 --note "review posted"
+fi
+
+# Who's working on what?
+twapp msg claim --list
+twapp msg claim --list --lane-prefix PR- --format json | jq
+
+# Stale-reclaim threshold (default 10 min). Match to your poll interval
+# so a busy peer isn't reclaimed mid-task.
+twapp msg claim audit-fees --stale-seconds 300
+```
+
+Only the current owner may release a lane. A claim whose `owner.json`
+is older than `--stale-seconds` and has no `released.json` is
+considered stale — any worker may force a re-claim, and the new owner
+records `reclaimed_from: <previous>` for the audit trail. See
+[`docs/designs/worker-coordination.md`](docs/designs/worker-coordination.md)
+for the full design (atomic-mkdir rationale, stale semantics, and
+what's out of scope).
+
 ## Spawning agent instances
 
 twapp works well as a terminal wrapper for *interactive* sessions, but it's
@@ -577,6 +613,9 @@ burns iteration.
 | `twapp msg send <to> [body]` | Send a message (writes to the shared mailbox inbox) |
 | `twapp msg broadcast [body]` | Broadcast to every handle (`to: [all]`) |
 | `twapp msg fetch [--for <h>] [--since <ts>]` | List inbox messages, filtered |
+| `twapp msg claim <lane-id> [--note <s>]` | Atomically claim a shared lane (PR, audit, backlog item); exit 1 if already claimed |
+| `twapp msg release <lane-id> [--note <s>]` | Release a lane you own; writes `released.json` and broadcasts the release |
+| `twapp msg claim --list [--lane-prefix <p>]` | List all active (unreleased, unstale) claims; `--format json` for machine-readable |
 | `twapp install-gui <binary>` | Install or update the app bundle |
 | `twapp setup-cert` | Create code signing certificate |
 | `twapp dev-reload --pid <pid>` | Rebuild and relaunch (dev workflow) |
--- a/docs/designs/worker-coordination.md
+++ b/docs/designs/worker-coordination.md
@@ -1,0 +1,212 @@
+# Design: Lane-claim coordination for N workers on a shared queue
+
+Status: **shipped** (initial primitive — `twapp msg claim / release`)
+Sibling doc: [`agent-messaging.md`](agent-messaging.md) — the broader
+filesystem-mailbox model this primitive plugs into.
+
+---
+
+## 0. Problem
+
+Two workers polling the same task queue will, given enough traffic, grab
+the same item at the same moment. Observed instance: two reviewer agents
+running simultaneously against the same PR list, both picking PR #91,
+both posting reviews, one's review arriving seconds after the other and
+fighting for the last-word.
+
+The pattern generalizes. Any N-worker-on-shared-queue scenario —
+reviewers on PRs, auditors on a backlog, implementers pulling from a
+prioritized list — needs a *claim* step before the *work* step, and a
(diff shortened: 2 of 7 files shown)


piekstra (commented):
**Ship it.** Solid coordination primitive, the atomic-mkdir resolver is the right shape, test coverage is dense including a real concurrent-race test, and the design doc + skill update land alongside the code per the coordinator's directive.

## Policy scan

Clean on the extended regex. Lane-id examples (`PR-91`, `audit-fees`, `backlog-item-7`, `PR-`) are generic.

## What's right

### The race resolver is correct

`std::fs::create_dir` is the right primitive. POSIX `mkdir(2)` is universally atomic, no seed file, no daemon, and no rename-races. The `concurrent_claims_produce_exactly_one_winner` test (8 threads gated on `std::sync::Barrier::wait`, then race) explicitly pins **1 winner / 7 losers** — that's the real load test, not a contrived one. ✓

### Three-state outcome model is the right abstraction

`Claimed` / `Reclaimed` / `AlreadyClaimed` map cleanly to:
- Different broadcast subjects (`claim lane X`, `reclaim lane X`, no broadcast on AlreadyClaimed).
- Different exit codes (0 / 0 / 1).
- Different audit-log shapes (presence/absence of `reclaimed_from` field).

The fact that `AlreadyClaimed` is returned (not errored) lets callers do `if twapp msg claim …; then …` without parsing stderr.

### Lane-id validation closes the obvious holes

`validate_lane_id` allows alnum + `- _ . : # @` only; rejects empty, `.`, `..`, length > 128, path separators, whitespace. Tested explicitly with `../escape`, `a/b`, `a\\b`, `a b` — all rejected. Path-traversal isn't reachable. ✓

(`owner/repo#123` is rejected because of the slash — that's correct for the on-disk storage; if a caller wants a GitHub-style identifier they need to swap `/` for something else like `_` or `:`. Worth one example in the README showing this transformation.)

### Atomic-write discipline

- **`owner.json`**: tmp-file with random suffix + rename. Concurrent reader can't observe a half-written file. ✓
- **`read_owner_with_wait`**: 20 × 10ms spin to wait out the create_dir-but-owner-not-yet-written window. Right call — race window is single-digit ms in practice; timeout of 200ms is generous without being absurd.
- **`archive_released_dir`**: rename with timestamp slug; falls back to random suffix on collision. Handles the `NotFound` race ("another worker already archived"). ✓

### Stale reclaim is honestly framed

The author calls it out explicitly: "not mkdir-atomic, but stale reclaims are coarse-grained coordination events that always emit a broadcast, so the audit trail catches any lost-race duplication." Worth one more line in the `try_claim` docstring to say so directly there too — readers won't always cross-reference the design doc.

### The design doc is well-shaped

`docs/designs/worker-coordination.md` has the right scope:
- §0 reproducer (two reviewer agents on PR #91 — exactly the failure today's coordinator-vs-twapp-reviewer-2 scenario rhymes with).
- §1 mechanics (claim → release → contest+reclaim) with a clear state table.
- §3 out-of-scope list: cross-host, crypto ownership, priority/preemption, byzantine workers, GitHub review-assignment, archive rotation. Each "we deliberately don't solve this" is named with reason. ✓
- §4 rationale section explains why mkdir > rename-lock > flock > GitHub review-assignment > database. Good defense against future "why didn't you just use X?" reviews.

### Skill updates land alongside the code per directive

- `agent-coordinator §8.5` gets a full N-worker-coordination section with reviewer-race example, stale-claim recovery, and briefing-template language. Single source of truth.
- `spawn-agent` gets a short cross-ref pointing at agent-coordinator. Right level — spawn-agent is per-spawn, agent-coordinator is per-fleet.

## Non-blocking nits

- **Dead code: `ClaimListFormat` + `From<ClaimListFormat> for FetchFormat`** at the bottom of `msg_claim.rs`. The actual `MsgCommands::Claim::format` is typed as `FetchFormat` (reusing #41's enum), not `ClaimListFormat`. The new enum and conversion impl are never reached. Drop them in a follow-up.
- **`age_seconds` returns 0 on RFC3339 parse failure**, which means a `claimed_at` value of `"garbage"` would never appear stale. An attacker (or a buggy peer) could permanently hold a lane by writing a malformed `owner.json`. Minor — the design's "filesystem ACLs are the trust boundary" already accepts this kind of risk — but a one-line `eprintln!("warn: lane {} has unparseable claimed_at", lane_id)` would surface the corruption to operators.
- **`released.json` is written non-atomically** (direct `std::fs::write`). For ~100-byte files most filesystems write atomically, but if you wanted to be belt-and-suspenders consistent with `owner.json`, the same tmp+rename pattern would work.
- **`agent-coordinator §8.5` numbering**: half-section numbers (8.5) are slightly awkward in a doc that otherwise uses whole numbers. Either renumber the whole doc or accept the half-section as a permanent insertion point. Non-blocking.
- **MSRV / env-var test pattern**: `MailboxGuard::new` calls `std::env::set_var` and `remove_var` without `unsafe { }`. PR #46 (merged, CI green on this MSRV) had the same shape — so this will compile today, but the future-proofing observation I made on #46 also applies here. If/when MSRV bumps to 1.82+, this needs the same `unsafe` wrap.
- **`README` "Lane claims" section** uses `--note` examples that are concrete ("reviewing", "review posted"). Worth a one-line note about what happens to the note in the broadcast subject vs body — the reader can infer from the example, but spelling it out helps.

## Coordination with other PRs in flight

- **No conflict with `MsgCommands` from #41** (merged) — `Claim`/`Release` are new variants on the existing enum.
- **No conflict with `Commands::Models` from #46** (merged) — top-level vs subcommand surface.
- **`skills/agent-coordinator/SKILL.md`**: this PR adds §8.5 to a section between §8 (stale-merge detection) and §9 (merge order). #43 (still open, currently DIRTY) and #47 (merged) also touch this file in different sections. #43's rebase will need to merge cleanly past §8.5 — should be straightforward since #43 modifies §2 template + §13 e2e example, neither of which overlaps with §8.5.
- **`docs/designs/worker-coordination.md`** is a brand-new file — no conflicts possible.

## Summary

Atomic mechanism is right, fail-safe paths are documented, the audit-trail-via-shadow-broadcast design is clever and uses existing #41 infrastructure rather than reinventing it. Test coverage is dense including the load-bearing 8-thread concurrent test. Skill + design-doc updates land in this same PR per the coordinator's standing directive.

**Ship it.** Tackle the dead-code and `age_seconds` log-on-malformed nits in a follow-up if/when. The coordinator can use this primitive immediately — exactly the situation that would have prevented today's `twapp-reviewer` / `twapp-reviewer-2` collision (the coordinator's own apology message anticipates exactly this).

PR #49, opened 2026-04-21
docs(agent-coordinator): add reap-sweep responsibility

piekstra:
## Summary

Add a standing §6.5 "Reap sweep" subsection to `skills/agent-coordinator/SKILL.md` that tells the coordinator to proactively stop scope-complete agents on every status sweep. Placed between §6 pre-merge governance and §7 stop signals to match the existing numbering pattern (§1b, §8.5).

Without this, zombie `twapp` hosts accumulate after their underlying claude agent has signed off or after a PR merges before the agent's /loop notices. The new section encodes when to run (every status check, post-merge, idle lull), the procedure (ps, cross-check PRs, verify stop condition), the ALL-must-hold reap conditions, the reap mechanism, and common mistakes — cross-linking §7's existing "Coordinator cleanup post-offboard" sub-list instead of duplicating it.

Docs-only. No `src/` or `src-tauri/` changes. No downstream renumbering required.

## Test plan

- [x] Scope compliance: only `skills/agent-coordinator/SKILL.md` touched
- [x] Cross-reference to §7 "Coordinator cleanup post-offboard" uses the skill's existing "(see §N)" style
- [x] Cross-reference to §13 for archetype lifespans to justify the long-running-role carve-out
- [x] No numbering churn in §7-§14

### DIFF
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -275,6 +275,60 @@ The reviewer answers "is this code correct and safe to run?" The coordinator ans
 
 Workers have self-merge authority (§5); this governance pass is the coordinator's counterweight. If the spot-check fails, post a block-message to the worker's mailbox before they merge — they will see it on their next poll.
 
+## 6.5. Reap sweep — proactively stopping scope-complete agents
+
+Run on **every status check** — any time the coordinator reports current state to the user, and whenever noticing an idle lull. Also run after any PR merge and on a direct user ask about agent status. Without a standing sweep, zombie `twapp` hosts pile up: PRs merge, `/loop`s keep polling, nothing offboards.
+
+### Procedure
+
+1. List running twapp hosts:
+
+   ```bash
+   ps -eo pid,pcpu,etime,command | grep "twapp --name"
+   ```
+
+2. For each running agent, resolve its scope:
+   - Does it have an open PR?
+   - Is it a long-running role (reviewer, log-watcher, auditor)?
+   - Is it the coordinator itself?
+
+3. Cross-reference against open PRs and the handle's declared stop condition:
+
+   ```bash
+   gh pr list --repo <repo> --state open
+   ```
+
+   Reread the briefing's `## Protocol` for its stop shape.
+
+4. **Reap only when ALL of these hold:**
+   - The agent's PR has merged or closed, AND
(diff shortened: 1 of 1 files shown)


piekstra (commented):

## Policy scan

## Correctness
- New §6.5 sits between §6 (pre-merge governance) and §7 (stop signals) — matches the §1b / §8.5 half-step convention already in use, so no downstream renumbering is needed and there's no broken anchor in the rest of the doc.
- The "reap only when ALL of these hold" list correctly encodes the three load-bearing conditions (PR merged/closed AND idle-or-offboarded AND not a long-running role), and the carve-out for reviewers / log-watchers / auditors points at §13 archetype lifespans which is the right cross-link.
- "Common mistakes" names the exact trap future coordinators will hit — treating 0% CPU as "dormant" when `/loop` is just between polls. Good.
- The "When in doubt, ask the agent" step gives you a bounded escape hatch (one poll interval) and avoids the much worse failure mode of killing a worker mid-push. Right instinct.

## Scope discipline
Docs-only, single file (`skills/agent-coordinator/SKILL.md`), +54 / -0. Matches the PR body exactly.

## Consistency
- Reap mechanism block cross-references §7 ("Coordinator cleanup post-offboard") instead of duplicating `twapp stop --name <handle>` + `git worktree remove`. Good — single source of truth.
- The "Forgetting the worktree cleanup" bullet is a direct pointer back at the §7 reminder, which is the right reinforcement for a step that's easy to skip.

- Step 4's "AND `/loop` has clearly gone idle (no mailbox posts in > `poll_interval × 3`, CPU 0.0%)" — `poll_interval × 3` is a heuristic that depends on the briefing's declared cadence. For the reviewer lane that's ~6 min, for slower lanes it's longer. Worth either naming a floor (e.g. "at minimum 5 min / 3 poll intervals, whichever is larger") or dropping the CPU 0.0% condition since, as the common-mistakes bullet notes, it's not diagnostic on its own. Borderline.
- "Optional — post a one-line audit broadcast for visibility. Not required." — consider making this the default for scope-complete reap (vs. idle-drain reap) so offboards stay visible to other handles. Matter of taste.
- The procedure's `ps -eo pid,pcpu,etime,command | grep "twapp --name"` would also match `grep "twapp --name"` itself when run from a shell that puts the grep into the process table; usually not an issue with `grep ... | grep -v grep` guarding but you omitted the `-v grep`. Minor shell hygiene.

Good addition. The previous gap — coordinator had no standing instruction to proactively reap zombies — was showing in practice (multiple `ps aux | grep "twapp --name"` piles noted in MORNING_GUIDE). This encodes the right protocol.

PR #50, opened 2026-04-21
feat(ui): message composer modal (send/broadcast)

piekstra:
## Summary

- Adds a modal in the session window (Actions ▸ **Send Message…** / **⌘⇧M**) that wraps `twapp msg send` / `twapp msg broadcast`. When `To` is `all` the body routes through `broadcast`; anything else goes through `send`.
- New Tauri command `send_message` shells out to the running `twapp` binary (via `std::env::current_exe()`), pipes the body through stdin (no argv length / shell-escape issues), parses the `Sent <path> (<id>)` stdout, and returns the 20-char id to the UI. Stderr is surfaced verbatim on CLI failure.
- Part of `docs/designs/agent-aware-ui.md` §5 — narrowed to the composer pane only per briefing. Inbox / broadcast / urgent / thread viewers and rich-text remain out of scope and land in wave-2 PRs.

## UX

- **Fields:** `To` (handle or `all`), `Priority` (routine/urgent/blocker, default routine), `Subject` (optional), `Thread` (disabled — reserved for a later Reply-to entry point per the briefing), `cc` (comma-separated), `Body` (plain markdown-ready textarea).
- **Validation:** recipient required, body required, priority must be routine/urgent/blocker. Errors render inline in the modal.
- **Keyboard:** `⌘⇧M` opens the modal, `Esc` closes, `⌘/Ctrl + Enter` submits from the body.
- **Discovery:** new **Send Message…** entry in the Actions dropdown next to Fork Session.
- **Feedback:** on success, a transient bottom-of-window toast `Message sent (id: <short>)` that auto-dismisses after 4s (or on click). On failure, inline modal error with the CLI's stderr.

## Test plan

- [x] `cargo test --manifest-path src-tauri/Cargo.toml --lib` — **127 passed**, includes 16 new in `gui::msg::tests`:
  - argv build: direct vs broadcast (incl. case-insensitive `all`), thread/cc dropped on broadcast, optional trimming, priority lowercased.
  - `parse_sent_id` — happy path, leading noise, missing-id returns None.
  - Validation — empty recipient / empty body / unknown priority / mixed-case priorities accepted.
  - Integration via a tempdir `fake-twapp` shell script: success path returns parsed id; failure path surfaces stderr contents.
- [x] `npm test` (vitest) — **85 passed**, includes 5 new for `validateArgs` (minimal valid, `all` accepted, empty recipient, empty body, optional blanks ignored).
- [x] `npm run dev` + `curl` smoke — Vite serves the component without a transform error.
- [ ] Manual GUI walkthrough (reviewer): `npm run tauri build && twapp install-gui ...` then exercise Actions ▸ Send Message, `⌘⇧M`, routine/urgent/blocker, `all` → broadcast, empty-recipient + empty-body validation banners, stderr surfacing (point `twapp` at an unreachable mailbox).

- Inbox viewer / message history (separate PR).
- Thread viewer / Reply-to prefill (depends on messaging PR-2 threading).
- Urgent-queue pane (wave-2, depends on priority-lane PR-4).
- Rich-text, attachments.

Briefing: `twapp-ui-msg-composer.md`. Coordinator handle: `twapp-ui-composer`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,6 +563,12 @@ twapp msg fetch --since 20260420T120000Z --limit 20
 twapp msg fetch --for reviewer --format json | jq
 ```
 
+Inside a twapp session window, the same `send` / `broadcast` actions are
+reachable from the UI: **Actions ▸ Send Message…** or the keyboard
+shortcut **⌘⇧M** (Ctrl+Shift+M on non-Mac) opens a composer modal that
+shells out to `twapp msg` under the hood. `To: all` routes through
+`broadcast`; any other value goes through `send`.
+
 If `--from` is not passed, the sender handle is taken from the current
 directory's `.twapp-session.json` `name`. Bodies may be passed as a
 positional argument or piped in on stdin.
--- a/src-tauri/src/gui/mod.rs
+++ b/src-tauri/src/gui/mod.rs
@@ -1,6 +1,7 @@
 pub mod config;
 pub mod files;
 pub mod monitor;
+pub mod msg;
 pub mod notes;
 pub mod prompts;
 pub mod pty;
@@ -111,6 +112,7 @@ pub fn run(args: GuiArgs) {
             config::set_monitor_float,
             monitor::list_monitor_logs,
             files::reveal_in_finder,
+            msg::send_message,
         ])
         .setup(move |app| {
             // Set window title — this controls the Mission Control fullscreen space label
--- a/src-tauri/src/gui/msg.rs
+++ b/src-tauri/src/gui/msg.rs
@@ -79,6 +79,10 @@ fn validate(args: &SendArgs) -> Result<(), String> {
 }
 
 fn twapp_binary() -> std::path::PathBuf {
+    // Contract: the Tauri GUI binary dispatches the CLI via subcommand (see
+    // `src-tauri/src/lib.rs`), so `current_exe()` is the same `twapp` binary
+    // that exposes `msg send` / `msg broadcast`. If the GUI is ever split out
+    // into its own binary, point this at a resolved `twapp` in PATH instead.
     std::env::current_exe().unwrap_or_else(|_| std::path::PathBuf::from("twapp"))
 }
 
@@ -278,10 +282,21 @@ mod tests {
         path
     }
 
+    fn unique_tmp(tag: &str) -> std::path::PathBuf {
+        use std::time::{SystemTime, UNIX_EPOCH};
+        let nanos = SystemTime::now()
+            .duration_since(UNIX_EPOCH)
+            .map(|d| d.as_nanos())
+            .unwrap_or(0);
+        let tmp = std::env::temp_dir().join(format!("twapp-send-{}-{}-{}", tag, std::process::id(), nanos));
+        let _ = std::fs::remove_dir_all(&tmp);
+        std::fs::create_dir_all(&tmp).unwrap();
+        tmp
+    }
+
     #[test]
     fn fake_cli_success_returns_id() {
-        let tmp = std::env::temp_dir().join(format!("twapp-send-ok-{}", std::process::id()));
-        std::fs::create_dir_all(&tmp).unwrap();
+        let tmp = unique_tmp("ok");
         let bin = write_fake_bin(
             &tmp,
             "fake-twapp-ok",
@@ -290,12 +305,12 @@ mod tests {
         let argv = build_argv(&base());
         let id = run_with_fake_bin(&bin, argv, "hello").expect("should succeed");
         assert_eq!(id, "FAKEID42");
+        let _ = std::fs::remove_dir_all(&tmp);
     }
 
     #[test]
     fn fake_cli_error_surfaces_stderr() {
-        let tmp = std::env::temp_dir().join(format!("twapp-send-err-{}", std::process::id()));
-        std::fs::create_dir_all(&tmp).unwrap();
+        let tmp = unique_tmp("err");
         let bin = write_fake_bin(
             &tmp,
             "fake-twapp-err",
@@ -304,5 +319,6 @@ mod tests {
         let argv = build_argv(&base());
         let err = run_with_fake_bin(&bin, argv, "hello").expect_err("should fail");
         assert!(err.contains("mailbox not found"), "expected stderr in error, got: {}", err);
+        let _ = std::fs::remove_dir_all(&tmp);
     }
 }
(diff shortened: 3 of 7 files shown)


piekstra (commented):

## Policy scan
Clean. No private-project terms hit the grep over the full diff. The word "broadcast" appears as mailbox-messaging vocabulary, not trading broker. Nothing to block on.

## Correctness
- `build_argv` is the right shape: single positional recipient for `send`, no recipient for `broadcast`, `--priority` always present, optional flags conditionally appended. The `is_broadcast` branch correctly drops `--thread` and `--cc` (broadcast doesn't accept them) and that's locked in by `argv_broadcast_drops_thread_and_cc`.
- stdin-piped body is the right call for long markdown / quoted content — avoids argv-length and shell-escape concerns. Good.
- `std::env::current_exe()` as the `twapp` binary assumes the GUI and CLI ship in the same binary (dispatch on subcommand). That's consistent with how the rest of the `gui/` module is structured and with your PR description, but worth noting as a contract: if `twapp-gui` ever splits into a separate binary, this breaks silently. A one-line comment on `twapp_binary()` to that effect would save a future reader 10 minutes of spelunking. Nit, not a blocker.
- `send_message` runs the child synchronously on the Tauri command thread. For a single inbox write that's fine — the CLI's `msg send` is fast. If a future variant ever hits the network (remote mailbox, etc.), move it off-thread; no action needed today.
- Frontend keyboard handler lives inside `if (!(e.metaKey || e.ctrlKey)) return;` — so the `if (e.key === "M" || (e.shiftKey && e.key === "m"))` pattern is guarded by the modifier check at the top of the handler. Mirrors Cmd+Shift+N. OK.

## Scope discipline
Files match the briefing: `src-tauri/src/gui/{mod,msg}.rs`, `src/App.{tsx,css}`, `src/components/MessageComposer.{tsx,test.ts}`, `README.md`. No unrelated drift. Inbox viewer / thread viewer / urgent queue / rich-text correctly deferred to wave-2 PRs per `docs/designs/agent-aware-ui.md` §5.

## Tests
16 Rust unit tests + 5 vitest tests cover argv build (including broadcast case-insensitivity, broadcast-drops-thread-cc, optional-whitespace trimming, priority lowercasing), `parse_sent_id` happy path + leading-noise + missing-id None, validation (empty recipient / empty body / unknown priority / mixed-case priority accepted), and a fake-bin integration test for both success id-parse and stderr surfacing. Good coverage for what's here.

Minor test hygiene nits (not blockers):
- `fake_cli_success_returns_id` + `fake_cli_error_surfaces_stderr` create dirs under `std::env::temp_dir().join(format!("twapp-send-...-{pid}"))` and never clean them up. On repeated runs with the same pid (rare but not impossible in CI), you could see stale state; `tempfile::tempdir()` gives you auto-cleanup for free.
- `MessageComposer.test.ts` doesn't exercise `validateArgs` with `priority` since validation is currently server-side only — worth a sentence or keep it as-is (the Rust side catches it).

## Docs consistency
README "Inside a twapp session window, the same `send` / `broadcast` actions are reachable…" block is in the right place (right after the `twapp msg` CLI section). Mentions `⌘⇧M` + Ctrl+Shift+M variant. `--help` output for `twapp msg` is unchanged, which is correct (no new subcommand on the CLI).

## UX nits (fold in only if cheap)
- The toast auto-dismisses at 4s via `setTimeout` but the previous timer isn't cleared when a second success arrives — rapid back-to-back sends will see the toast clear at the earlier-scheduled time. Minor.
- The `Send Message...` Actions entry and the `⌘⇧M` shortcut aren't cross-advertised in the UI (neither the button nor the menu item shows the shortcut). Low cost to add `<span className="shortcut-hint">⌘⇧M</span>` alongside the button label.
- Manual GUI walkthrough is flagged incomplete in the PR body. Before merge, one end-to-end `npm run tauri build` round with Actions ▸ Send Message, `⌘⇧M`, routine/urgent/blocker, `all` → broadcast, and an induced stderr surface is worth the 5 min — Tauri command IPC has a habit of surfacing shape mismatches that unit tests miss.

Nothing above is a block. With CI green and the unit + integration suite clean, this is ready to merge once you've done the manual smoke.

piekstra (commented):
**Ship it.** Clean composer modal with solid argv/parse plumbing and real integration tests.

## Policy scan

Clean on the extended regex.

## What's right

- **`build_argv` is a pure function** with 7 focused unit tests covering direct vs broadcast routing, case-insensitivity on `all`, thread/cc drop on broadcast, optional field skipping, priority normalization, and empty-field-filter. The shape pins the CLI contract before any I/O.
- **Body via stdin, not argv** — no shell-escape / argv-length risk for long markdown bodies. The right call.
- **`twapp_binary()` uses `current_exe()`** — the GUI binary dispatches the CLI subcommand in `lib.rs`, so the same process image has both. The docstring explicitly names this contract, with a migration pointer if the binary ever splits. Clean.
- **Integration tests with fake shell scripts** (`fake-twapp-ok`, `fake-twapp-err`) verify both the success-id-parse path and the stderr-passthrough path. That's the right level of integration coverage for a shell-out — unit-testing the argv shape + mocking the CLI interface end-to-end.
- **`parse_sent_id` handles leading noise**, missing id, empty stdout. The `rfind('(')` + `rfind(')')` approach is robust against subject strings containing parens.
- **Validation**: recipient required, body required, priority must be in the fixed set. Priority accepted case-insensitively (`Routine`/`URGENT`/`blocker` all pass) but normalized to lowercase for the CLI argv. ✓
- **UX matches design §5**: modal with To/Priority/Subject/Thread (stubbed)/cc/Body, ⌘⇧M shortcut, Esc closes, ⌘/Ctrl+Enter sends. Thread field disabled with a note — right call; wire the reply-to entry point in a later wave.
- **Scope explicitly narrowed** in the PR body: "narrowed to the composer pane only per briefing. Inbox / broadcast / urgent / thread viewers and rich-text remain out of scope and land in wave-2 PRs." ✓

## Non-blocking nits

- **`send_message` swallows `config.cwd` if `None`**. If a user opens the modal outside a session window, the CLI's `--from` fallback to `.twapp-session.json::name` will fail with a clear error. That's fine — the modal is only reachable from a session window today — but worth a docstring comment on the Tauri command explaining the cwd dependency.
- **`parse_sent_id` uses `rfind`** on `(` and `)`. A subject containing `(...)` would be correctly skipped because the id is always the last parenthetical. ✓ But if the CLI ever wraps its stdout line (line-wrap), this would break. Not a concern today — the format is stable.
- **Integration tests are UNIX-only** (`#[cfg(unix)]` on the `chmod` step). Windows runners would skip them. Acceptable — twapp is darwin-first per the environment info.

## Summary

Right scope per design doc §5, right plumbing (stdin, pure argv builder, clear CLI contract), right test coverage (unit + shell-out integration). CI green.

**Ship it.**

PR #51, opened 2026-04-21
feat(msg): archive rotation + purge — design PR-7

piekstra:
## Summary

PR-7 of `docs/designs/agent-messaging.md` — daily archive maintenance. Adds `twapp msg archive {rotate,purge,list}`.

- **`rotate`** moves flat `<mailbox>/archive/*.md` into `archive/<YYYY-MM-DD>/`. Date is read from the message's frontmatter `ts` first, falls back to the filename's `YYYYMMDDTHHMMSSZ` prefix, then to the file's mtime. Idempotent — a second run on a already-rotated archive finds nothing to move.
- **`purge [--retain-days N] [--dry-run]`** removes day-directories older than N (default 14). Never touches `inbox/`, `presence/`, `cursors/`, `claims/`, or the current day.
- **`list [--since <YYYY-MM-DD>] [--format json]`** reports archived message counts per day.

Cron-friendly: every subcommand exits 0 on success or no-op and non-zero only on filesystem errors.

## Docs

- README grows an **Archive maintenance** subsection with an example daily cron line.
- `skills/agent-coordinator/SKILL.md` §3 gains a one-line mention that coordinators may run `twapp msg archive rotate` daily.

## Test plan

- [x] `cargo test --lib` — 128 tests pass (17 new in `msg_archive`, 111 pre-existing).
- [x] Required tests per the briefing: `rotate_moves_old_messages_into_date_subdirs`, `rotate_is_idempotent`, `rotate_prefers_frontmatter_ts_over_mtime`, `purge_removes_archives_older_than_retain_days`, `purge_dry_run_does_not_delete`, `purge_preserves_current_day`, `list_counts_by_day`.
- [x] Extra coverage: bare-legacy filename fallback, rotate dry-run, rotate leaves existing date subdirs alone, purge skips non-date dirs, purge on missing archive is a no-op, list honors `--since`, list counts nested subdirs, strict `YYYY-MM-DD` date-dir parser.
- [x] Smoke test against `/tmp` mailbox: `list` → `rotate --dry-run` → `rotate` → `list --format json` → second `rotate` (no-op).

- `cursors/` retention (PR-3 / PR-5).
- Encryption / compression of archived messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -612,6 +612,43 @@ records `reclaimed_from: <previous>` for the audit trail. See
 for the full design (atomic-mkdir rationale, stale semantics, and
 what's out of scope).
 
+#### Archive maintenance
+
+After workers archive their read messages (see
+[`skills/agent-coordinator/SKILL.md`](skills/agent-coordinator/SKILL.md#archiving)),
+the flat `<mailbox>/archive/` fills up over time. `twapp msg archive`
+provides three cron-friendly subcommands for keeping it tidy:
+
+```bash
+# Move flat archive/*.md messages into archive/<YYYY-MM-DD>/ by
+# frontmatter ts (falling back to filename ts, then file mtime).
+# Idempotent — run as often as you like.
+twapp msg archive rotate
+twapp msg archive rotate --dry-run     # show planned moves, touch nothing
+
+# Drop day-directories older than --retain-days (default 14).
+# Never touches inbox/, presence/, cursors/, claims/, or the current day.
+twapp msg archive purge
+twapp msg archive purge --retain-days 30
+twapp msg archive purge --dry-run
+
+# Per-day message counts.
+twapp msg archive list
+twapp msg archive list --since 2026-04-01
+twapp msg archive list --format json | jq
+```
+
+All three exit 0 on success or no-op and non-zero only on filesystem
+errors, so they're safe to wire into a daily cron:
+
+```cron
+# Daily at 04:15 — rotate yesterday's flat files, then purge >14 days.
+15 4 * * * TWAPP_MAILBOX_DIR=$HOME/collab/mailbox /usr/local/bin/twapp msg archive rotate && TWAPP_MAILBOX_DIR=$HOME/collab/mailbox /usr/local/bin/twapp msg archive purge
+```
+
+See [`docs/designs/agent-messaging.md` §2.8](docs/designs/agent-messaging.md)
+for the retention rationale.
+
 ### Roles and provenance
 
 Every session carries two extra pieces of metadata on
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -179,6 +179,10 @@ mv <shared-dir>/mailbox/inbox/<msg>.md <shared-dir>/mailbox/archive/
 
 Messages addressed to *other* agents stay in `inbox/` — never archive on someone else's behalf. The coordinator sweeps stragglers periodically (agents that offboarded without tidying).
 
+### Archive maintenance
+
+The flat `archive/` fills up over time. Coordinators may run `twapp msg archive rotate` daily to partition it into `archive/<YYYY-MM-DD>/` dirs, and `twapp msg archive purge` to drop days older than 14. See the README "Archive maintenance" section for the cron line.
+
 ### Addressing
 
 - `to: all` — broadcast (every online agent reads and archives).
--- a/src-tauri/src/cli/mod.rs
+++ b/src-tauri/src/cli/mod.rs
@@ -4,6 +4,7 @@ pub mod coordinator;
 pub mod models;
 pub mod monitor;
 pub mod msg;
+pub mod msg_archive;
 pub mod msg_claim;
 pub mod notes;
 pub mod permissions;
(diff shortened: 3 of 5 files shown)


piekstra (commented):

## Policy scan
Clean. No private-project terms. "broadcast" and "direct" appear only as archive sub-bucket names inside the existing messaging design.

## Correctness

### `rotate_archive`
- Precedence is right: frontmatter `ts` → filename `YYYYMMDDTHHMMSSZ` prefix → file mtime. `rotate_prefers_frontmatter_ts_over_mtime` locks in the "filename lies but frontmatter tells the truth" case, which is the one that matters for hand-edited or re-timestamped files.
- Top-level-only scan (`if !file_type.is_file() { continue; }`) means already-nested `archive/<date>/...` files are not re-rotated. `rotate_leaves_existing_date_subdirs_untouched` nails this. Idempotent per `rotate_is_idempotent`.
- `std::fs::rename` across dir boundaries within the same filesystem is atomic on Unix; across mount boundaries it'll fail with EXDEV and the error surfaces. Acceptable for PR-7; worth a follow-up note only if anyone ever mounts `archive/` on a separate filesystem.

### `purge_archive`
- Strict `parse_date_dir` (10-char digit check before `chrono::NaiveDate::parse_from_str`) correctly rejects `2026-4-18` and `2026-13-01`. Without it, chrono's `%m`/`%d` would silently accept lenient forms — explicit comment in the code documents that. Good.
- Current-day preservation (`if name == today_str { continue; }`) runs before the cutoff check, so `retain_days=0` still keeps today. `purge_preserves_current_day` pins this invariant — it's the safety rail you'd regret removing later.
- Non-date directories (`scratch/`, a coordinator dir, etc.) are silently skipped, not errored on. Correct disposition for a cron-friendly tool.
- Injected `today: chrono::NaiveDate` keeps the boundary deterministic in tests, production passes `Utc::now().date_naive()`. Right pattern.

### `list_archive`
- `count_md_files` walks recursively, matches `.md` case-insensitively, accumulates via an explicit stack (no recursion depth concern). `list_counts_nested_subdirs` pins the `broadcast/`, `direct/reviewer/` case from design §2.8.
- Results are sorted ascending by date string — lexicographic ordering is fine here because all names are `YYYY-MM-DD` padded.
- `--since` comparisons use `NaiveDate`, not string compare. Good.
- `cmd_list` handles a serde_json error, which is structurally unreachable for `Vec<DayCount>` (String + usize). Harmless dead branch.

### CLI surface
- `twapp msg archive rotate|purge|list` subcommands wired cleanly through `MsgCommands::Archive { command }` in `cli/mod.rs` and `cli/msg.rs`. The `#[command(after_help = "Examples:\n  ...")]` on each leaf gives useful `--help` output.
- Exit-code contract per subcommand (`0` on success or no-op; non-zero only on FS errors) matches the PR body and is cron-safe. `cmd_rotate` / `cmd_purge` / `cmd_list` all return `i32` uniformly.

## Scope discipline
Files match the briefing exactly: new `src-tauri/src/cli/msg_archive.rs` (792), touchups in `cli/mod.rs` (+13) + `cli/msg.rs` (+10 for the `Archive` subcommand), README `Archive maintenance` section (+37), and one line in the coordinator skill pointing at it (+4). `inbox/`, `presence/`, `cursors/`, `claims/` deliberately untouched — correct.

## Tests
17 new, all the briefing-mandated ones (`rotate_moves_old_messages_into_date_subdirs`, `rotate_is_idempotent`, `rotate_prefers_frontmatter_ts_over_mtime`, `purge_removes_archives_older_than_retain_days`, `purge_dry_run_does_not_delete`, `purge_preserves_current_day`, `list_counts_by_day`) plus seven extras (bare-legacy filename fallback, rotate dry-run, `rotate_leaves_existing_date_subdirs_untouched`, `purge_ignores_non_date_directories`, `purge_on_missing_archive_is_no_op`, `list_counts_nested_subdirs`, `list_on_missing_archive_is_empty`, `date_from_compact_ts_parses_or_rejects`, `parse_date_dir_accepts_only_well_formed_names`). `TempArchive` with `uuid::Uuid::new_v4()` + `Drop` cleanup is the right pattern — no stale state between runs.

## Docs consistency
- README "Archive maintenance" block slots correctly under the existing `twapp msg` section and is discoverable next to `fetch` / `ack`. Example cron line is copy-pasteable and uses `TWAPP_MAILBOX_DIR` (matches `resolve_mailbox_dir()`). Good.
- Skill §3 one-liner pointer is right-sized — avoids duplicating the README prose.

- `rotate_archive` builds the full move plan even in non-dry-run mode before doing any filesystem work. If a `rename` fails partway through, earlier files have already been moved. A recovery note in the docs ("partial rotation is safe to retry — idempotent") would eliminate future-worry about interrupted runs; the behavior is already correct, just worth making explicit.
- `parse_date_dir("2026-13-01")` relies on chrono rejecting month=13. It does today; worth a quick comment next to the test so a future refactorer doesn't replace it with a lenient parser.
- Cron example pins `TWAPP_MAILBOX_DIR` twice on the same line. Users who forget to re-export between the `&&` will see `rotate` work and `purge` silently target a different mailbox. A one-liner in the doc warning against that (or wrapping in `env VAR=... bash -c 'rotate && purge'`) would harden it against foot-guns. Minor.

Clean PR. Straightforward maintenance CLI, good test discipline, cron-safe exit codes.

PR #53, opened 2026-04-21
feat(msg): priority lane (urgent/ symlinks + --priority filter) — design PR-4

piekstra:
## Summary

Implements **PR-4** of the [agent-messaging design](docs/designs/agent-messaging.md) (§2.5, §5).

- `twapp msg send --priority {urgent|blocker}` (and `broadcast`) write the canonical file under the flat `inbox/` **and** create a symlink under `inbox/urgent/<recipient>/<ts>-<id6>.md` → canonical path.
  - Broadcasts → `inbox/urgent/all/`.
  - Direct to N recipients → one symlink per recipient.
  - `channel:` recipients are skipped (land in PR-6).
  - Routine traffic never touches the urgent lane.
- `twapp msg fetch --priority {urgent|blocker}` takes a fast path: scans `inbox/urgent/<for-handle>/` + `inbox/urgent/all/` instead of the flat inbox, then frontmatter-filters.
  - `--priority urgent` returns the whole urgent lane (urgent + blocker).
  - `--priority blocker` is exact.
  - `--priority routine` is exact.
- Broken symlinks (canonical deleted, lane entry not swept yet) are skipped with a `log::debug!` trace, not a crash.

## Docs

- README: new "Priority lanes" subsection under "Messaging between sessions".
- `skills/agent-coordinator/SKILL.md` §3: `--priority blocker` idiom for time-sensitive coordinator asks, and a deprecation line noting that the `-URGENT` / `-REWORK` filename-hack idiom is replaced by `--priority urgent`.

## Test plan

All 6 tests from the briefing + one bonus for the broadcast path:

- [x] `send_urgent_creates_symlink_in_urgent_lane`
- [x] `send_routine_does_not_create_urgent_symlink`
- [x] `fetch_priority_urgent_returns_urgent_and_blocker_not_routine`
- [x] `fetch_priority_blocker_returns_only_blocker`
- [x] `broken_symlink_in_urgent_lane_does_not_crash_fetch`
- [x] `multi_recipient_direct_urgent_symlinks_one_per_recipient`
- [x] `broadcast_urgent_lands_in_urgent_all` (bonus)
- [x] Full test suite: 118 passed / 0 failed.
- [x] E2E smoke against a temp mailbox: urgent direct + blocker broadcast symlink correctly; routine does not; `fetch --priority urgent --for reviewer` returns both the direct urgent and the broadcast blocker via `urgent/all/`.

## Out of scope

- Presence (PR-5), channels (PR-6), directory split (PR-3). Channel recipients are deliberately skipped from urgent-lane symlinking; that wiring lands with PR-6.

### DIFF
--- a/README.md
+++ b/README.md
@@ -519,10 +519,12 @@ without leaving twapp.
 fenced-frontmatter markdown files into a shared `inbox/` directory and
 reads them back. See
 [`docs/designs/agent-messaging.md`](docs/designs/agent-messaging.md) for
-the full model. Today's CLI covers **PR-1** of that design — `send`,
-`broadcast`, and `fetch` against the current flat `inbox/` layout.
-Threading, directory split, cursors, priority lanes, presence, channels,
-and archive rotation all land in later PRs.
+the full model. Today's CLI covers **PR-1** (scaffolding) and **PR-4**
+(priority lane) of that design — `send`, `broadcast`, and `fetch`
+against the current flat `inbox/` layout, plus urgent-lane symlinks
+under `inbox/urgent/<recipient>/` for time-sensitive traffic.
+Threading, directory split, cursors, presence, channels, and archive
+rotation all land in later PRs.
 
 #### Configure the mailbox
 
@@ -566,6 +568,41 @@ twapp msg fetch --for reviewer --format json | jq
 If `--from` is not passed, the sender handle is taken from the current
 directory's `.twapp-session.json` `name`. Bodies may be passed as a
 positional argument or piped in on stdin.
+
+#### Priority lanes
+
+Messages carry a `priority:` frontmatter field — `routine` (default),
+`urgent`, or `blocker`. When a message is sent with `--priority urgent`
+or `--priority blocker`, `twapp msg send` (and `broadcast`) additionally
+writes a symlink under `inbox/urgent/<recipient>/<ts>-<id6>.md` pointing
+at the canonical file in the flat inbox. Broadcasts symlink into
+`inbox/urgent/all/`; multi-recipient direct messages get one symlink
+per recipient.
+
+```bash
+# Urgent — asks the recipient to interrupt their batch and read next poll.
+twapp msg send reviewer --priority urgent --subject "scope change" "see body"
+
+# Blocker — asks the recipient to stop current work and handle this first.
+twapp msg send worker-a --priority blocker --subject "rewind PR" "force-push coming"
+
+# Fetch only what deserves attention. `--priority urgent` is the "urgent
+# lane" — it returns both urgent AND blocker traffic. `--priority blocker`
+# is an exact match for the rare stop-the-world case.
+twapp msg fetch --priority urgent            # urgent + blocker
+twapp msg fetch --priority blocker           # blocker only
+twapp msg fetch --priority routine           # everything else
+```
+
+Under the hood, `fetch --priority urgent|blocker` scans
+`inbox/urgent/<you>/` (and `inbox/urgent/all/` for broadcasts) first —
+much cheaper than listing the flat inbox as it grows. Routine traffic
+stays out of the urgent lane entirely. Broken symlinks (canonical
+deleted, lane entry not yet swept) are skipped with a `log::debug!`
+trace, not a crash.
+
+> `twapp msg fetch --priority blocker` is the idiomatic first call at the
+> top of any long write cycle — before the first line of code.
 
 #### Reading legacy (bare) files
 
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -185,6 +185,33 @@ Messages addressed to *other* agents stay in `inbox/` — never archive on someo
 - `to: <handle>` — directed (only that agent archives; others ignore).
 - `cc: <handle>` — optional courtesy copy.
 
+### Priority lanes (urgent / blocker)
+
+When a coordinator message is time-sensitive (redirect mid-write-cycle,
+stop-work order, stale-merge nudge), send it on the priority lane so
+the recipient surfaces it on its next poll without scanning the whole
+inbox:
+
+```bash
+# "Read this on your next poll; it may change your plan."
+twapp msg send worker-a --priority urgent --subject "scope change" "<body>"
+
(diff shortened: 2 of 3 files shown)


piekstra (commented):

## Policy scan
Clean. No private-project terms. "broadcast", "urgent", "blocker" are mailbox-messaging vocabulary per design §2.5.

## Correctness

### Write path (`create_urgent_symlinks`)
- Only fires on `Urgent | Blocker` — routine stays out of the lane entirely. `send_routine_does_not_create_urgent_symlink` additionally asserts no per-recipient `urgent/reviewer/` dir is created, which is the right thing (keeps `ls inbox/urgent/` tidy).
- `channel:` recipients are correctly skipped — channel-lane wiring is PR-6 scope.
- Relative symlink target (`../..` + filename) means moving the mailbox tree doesn't break links. Good.
- Idempotent: `if link_path.exists() || link_path.is_symlink() { continue; }` before `std::os::unix::fs::symlink`. Two writers racing to the same id6 (astronomically unlikely — 20-char Crockford base32) won't error.
- Error disposition is correct for a best-effort secondary lane: `create_dir_all` or `symlink` failures log at `debug!` and move on — the canonical message is already on disk, so the lane is strictly additive. A reader that doesn't know about the urgent lane still sees the message via the flat inbox. No silent data loss.
- `unix::fs::symlink` only — no Windows fallback. Matches the rest of twapp (Unix-only Tauri target). Fine.

### Read path (`scan_urgent_lane` + `select_messages`)
- Per-handle: scans `urgent/<handle>/` AND `urgent/all/` unless `handle == "all"`, in which case it skips the redundant append. Correct broadcast semantics — `broadcast_urgent_lands_in_urgent_all` covers it.
- `None` handle scans every dir under `urgent/`. The `.filter(|p| p.is_dir())` guard skips any stray files the user dropped there by hand. Good.
- Broken-symlink disposition: `std::fs::canonicalize` failure is logged at `debug!` and skipped; `broken_symlink_in_urgent_lane_does_not_crash_fetch` pins both the "canonical deleted after link" case and the "someone manually put a ghost link in urgent/" case. Good coverage.
- De-dup via canonical-path `HashSet` — if a recipient is in the direct lane AND a broadcast `urgent/all/` link points at the same canonical file, they only see it once. Subtle but important.
- Sort by `(fm.ts, fm.id)` — stable ordering even when two messages share a second-precision timestamp. Good.

### Filter semantics (`select_messages`)
- `--priority urgent` → `urgent | blocker` (the whole "deserves attention" lane); `--priority blocker` → exact-match blocker only; `--priority routine` → exact-match routine. Matches how a reader thinks about it, and the README line explains the distinction clearly ("`fetch --priority urgent` is the urgent lane"). `fetch_priority_urgent_returns_urgent_and_blocker_not_routine` + `fetch_priority_blocker_returns_only_blocker` nail both branches.
- Correctly routes the scan: urgent/blocker → lane fast path; routine / None → flat `list_messages`. The fast path is meaningful once the flat inbox grows — directly avoids the FS scan cost.
- Filter pipeline order is unchanged from the inlined version in `cmd_fetch`, just factored out. Good refactor.

### CLI refactor
- `cmd_fetch` now calls `select_messages(&inbox, …)` with the same args it used to retain inline. Zero behavior change on the non-priority paths — the extraction is mechanical.

## Scope discipline
Files match the briefing: `src-tauri/src/cli/msg.rs` (+393/-23), `README.md` (Priority lanes subsection + banner update), `skills/agent-coordinator/SKILL.md` (Priority lanes subsection + deprecation of `<handle>-URGENT` / `<handle>-REWORK` filename idiom). Nothing drifting into other modules, nothing in PR-3 / PR-5 / PR-6 territory.

## Tests
All 6 briefing-required plus one bonus (`broadcast_urgent_lands_in_urgent_all`). 118 total pass. Notable coverage:
- `fetch_priority_urgent_returns_urgent_and_blocker_not_routine` — the core lane-semantics test
- `broken_symlink_in_urgent_lane_does_not_crash_fetch` with BOTH the "I deleted the canonical after sending" case AND a hand-crafted orphan link — exercises the two real-world failure modes.
- `multi_recipient_direct_urgent_symlinks_one_per_recipient` checks every recipient's fetch returns exactly one message, not duplicates — the de-dup seen-set is load-bearing here.

One nit: the sleeps (`std::thread::sleep(Duration::from_millis(1100))`) between sends to force distinct `ts` seconds will add ~4.4s to the test run when collected together. For the filter-semantics tests, consider bumping the ts-granularity downstream or stubbing the clock in `now_compact()` instead. Not a blocker.

## Docs consistency
- README's "Messaging between sessions" banner updates from "PR-1 only" to "PR-1 + PR-4", with a proper "Priority lanes" subsection giving the three idiomatic invocations (urgent for send, blocker for the rare case, urgent for fetch returning urgent+blocker). The callout `> twapp msg fetch --priority blocker is the idiomatic first call at the top of any long write cycle` is exactly the norm you want agents to internalize.
- Coordinator skill §3's deprecation of the `<handle>-URGENT` / `<handle>-REWORK` filename hack is overdue — that workaround has been showing up in the archive for weeks. Good cleanup.
- `--help` output for `twapp msg send` / `broadcast` / `fetch` is unchanged (same `--priority` flag, same values). No CLI surface change for callers that weren't using the new lane. Correct.

- `create_urgent_symlinks` silently swallows the "urgent dir create failed" case and returns after logging. If the caller (`write_message`) wants to report partial success, the return shape today can't express it. Acceptable for PR-4, but note that a `Result<Vec<PathBuf>, String>` return + best-effort wrapper at the call site would give observability for the rare disk-error case. Follow-up.
- `scan_urgent_lane` on `None` handle does a single-level `read_dir` and doesn't recurse into sub-subdirs. That's correct for the current flat layout, but the "channel:" comment in the symlink writer implies sub-hierarchy could appear later. Worth a `debug_assert!` or a comment pinning the invariant.
- The `#[allow(clippy::too_many_arguments)]` on `select_messages` is fine, but a struct (`FetchFilter { for_handle, since, priority, thread, channel, limit }`) would scale better when PR-3 / PR-5 add more filters. Also follow-up.

Clean PR. The symlink-based urgent lane is the right lightweight design for PR-4 — no schema migration, readers can opt in gradually, deprecated filename idiom is cleanly retired.

PR #54, opened 2026-04-21
feat: session colab_group field + auto-inheritance from parent session

piekstra:
## Summary

Adds a `colab_group: Option<String>` field on `.twapp-session.json`
alongside the existing `role` / `provenance` metadata, plus the CLI
ergonomics to populate it.

- **`SessionData`**: new serde-default + `skip_serializing_if = Option::is_none`
  field. Legacy session files round-trip unchanged.
- **`twapp work --colab-group <name>`**: explicit flag (empty rejected).
  When `--from-file` is set and the flag is omitted, auto-inherits from
  the spawning session by walking upward from cwd looking for a
  `.twapp-session.json`. Direct `twapp work` (no `--from-file`) never
  auto-inherits — ad-hoc sessions stay ungrouped by default.
- **`twapp coordinator launch [--colab-group <name>]`**: default is
  `--name` so workers spawned from the coordinator's session auto-collect
  into a group named after the coordinator.
- **`twapp coordinator claim [--colab-group <name>]`**: set/overwrite
  `colab_group` at claim time; omit to leave it untouched.
- **CLI + GUI fork paths** inherit `colab_group` from the parent session
  (`read_fork_inherited_metadata` now returns a 3-tuple).
- **`twapp sessions` output** gains a `Colab` column: `colab=<group>` or `-`.
- **Docs**: `skills/agent-coordinator/SKILL.md` §1b, `skills/spawn-agent/SKILL.md`
  `--colab-group` bullet, README co-lab grouping subsection.

## Tests

All 8 briefing-required tests land, plus 8 supporting covering edge cases
(empty-string rejection, whitespace-only parent value, column formatting,
claim colab updates on already-coordinator sessions). Full `cargo test --lib`:
**127 passed / 0 failed** (up from 83).

Briefing-required tests:

- [x] `session_serde_roundtrip_with_colab_group`
- [x] `session_serde_roundtrip_without_colab_group`
- [x] `work_colab_group_flag_sets_field`
- [x] `coordinator_launch_defaults_colab_group_to_name`
- [x] `coordinator_launch_colab_group_override`
- [x] `from_file_inherits_colab_group_from_parent_session`
- [x] `from_file_without_parent_colab_group_leaves_field_unset`
- [x] `empty_colab_group_errors`

## Coordinates with

`twapp-ui-session-groups` (parallel consumer of this schema). Shipping
this first unblocks that worker to consume `SessionData.colab_group`
directly.

## Test plan

- [x] `cargo test --lib` — 127/127 passing
- [x] `cargo check --bin twapp` — clean
- [x] Briefing's 8 tests all present by the exact names listed above
- [x] Legacy `.twapp-session.json` (no `colab_group` key) deserializes cleanly
- [x] `twapp sessions` plain output renders `colab=<group>` / `-` without
      widening enough to blow rows for long group names (truncated at 13 chars)

## Out of scope

- UI reorg / "My sessions" vs per-group sections — that's `twapp-ui-session-groups`.
- Per-group shared mailbox dir — separate future concern; this PR only
  adds the identifier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -642,6 +642,44 @@ workers and supervisors are visually distinct. Legacy
 `.twapp-session.json` files without these fields continue to load
 unchanged.
 
+### Co-lab groups
+
+When a coordinator supervises several workers, group membership is
+tracked by the `colab_group` field on `.twapp-session.json`. The
+convention is that a coordinator's group name equals its `--name`, and
+workers spawned under that coordinator auto-inherit the group. That
+lets the launcher render "My sessions" vs per-coordinator buckets
+without any extra coordination overhead.
+
+- `twapp coordinator launch --name infra-coord` stamps
+  `colab_group = "infra-coord"`. Override with
+  `--colab-group <name>` when the group should differ from the
+  coordinator's display name.
+- `twapp work --from-file <briefing>` with no explicit
+  `--colab-group` walks upward from the calling shell's cwd looking for
+  a `.twapp-session.json` and inherits the parent's `colab_group` when
+  one is set. User-typed `twapp work` invocations without `--from-file`
+  never inherit — ad-hoc sessions stay ungrouped by default.
+- `twapp coordinator claim --colab-group <name>` sets or overwrites the
+  group on an existing session (handy when you realize mid-flight that
+  a session belongs to a co-lab).
+
+```bash
+# Coordinator anchors the group.
+twapp coordinator launch --name feature-x   # colab_group="feature-x"
+
+# Worker auto-inherits when spawned via --from-file.
+twapp work --name worker-a --from-file /path/to/worker-a.md
+
+# Opt-out / override an inherited group.
+twapp work --name audit-helper --from-file /path/brief.md --colab-group ""  # rejected
+twapp work --name audit-helper --from-file /path/brief.md --colab-group shared-audit
+```
+
+`twapp sessions` renders a `Colab` column (`colab=<group>`) alongside
+the existing `Role` column. Sessions without a group show `-`; legacy
+session files without `colab_group` load unchanged.
+
 ### Model selection per agent
 
 `twapp work --model <name>` pass-through sets the model on the spawned
@@ -767,8 +805,8 @@ burns iteration.
 | `twapp msg claim <lane-id> [--note <s>]` | Atomically claim a shared lane (PR, audit, backlog item); exit 1 if already claimed |
 | `twapp msg release <lane-id> [--note <s>]` | Release a lane you own; writes `released.json` and broadcasts the release |
 | `twapp msg claim --list [--lane-prefix <p>]` | List all active (unreleased, unstale) claims; `--format json` for machine-readable |
-| `twapp coordinator launch [--briefing <p>] [--name <n>] [--shared-dir <d>]` | Spawn a fresh session wired as coordinator (writes `role: "coordinator"`) |
-| `twapp coordinator claim [--name <n>] [--force]` | Re-tag an existing session's role to `coordinator` |
+| `twapp coordinator launch [--briefing <p>] [--name <n>] [--shared-dir <d>] [--colab-group <g>]` | Spawn a fresh session wired as coordinator (writes `role: "coordinator"`; default `colab_group = --name`) |
+| `twapp coordinator claim [--name <n>] [--force] [--colab-group <g>]` | Re-tag an existing session's role to `coordinator` (optionally set/overwrite `colab_group`) |
 | `twapp install-gui <binary>` | Install or update the app bundle |
 | `twapp setup-cert` | Create code signing certificate |
 | `twapp dev-reload --pid <pid>` | Rebuild and relaunch (dev workflow) |
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -57,10 +57,26 @@ it tells the new coordinator to read this skill, register in
 `handle.txt`, post a hello, and start `/loop` polling. Override it with
 `--briefing <path>` whenever project-specific scope needs to ride along.
 
+`launch` also auto-assigns the session's `colab_group` to the
+coordinator's `--name` (so by default all workers it spawns from that
+session — via `twapp work --from-file` — auto-inherit the same group and
+group cleanly under the coordinator in `twapp sessions` / launcher UI).
+Override with `--colab-group <name>` when the group name should differ
+from the coordinator's display name (for example, when one coordinator
+runs multiple disjoint co-labs back-to-back).
+
+```bash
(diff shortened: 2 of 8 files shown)


piekstra (commented):

## Policy scan
Clean. No private-project terms. "colab" appears as the new field name; "coordinator" / "worker" are already established mailbox-vocabulary. Nothing to flag.

## Correctness

### Schema (`SessionData.colab_group`)
- `#[serde(default, skip_serializing_if = "Option::is_none")]` is the right idiom for additive schema growth: legacy `.twapp-session.json` without the field deserializes to `None`; sessions without a group round-trip without polluting the JSON. `session_serde_roundtrip_with_colab_group` and `..._without_colab_group` lock the invariant in both directions.
- Documented right at the field: "Convention is the coordinator's `--name`, but any free-form string is allowed." Matches both the UI partitioner (PR #52) and the auto-inheritance heuristic. Good.

### `resolve_colab_group` — CLI flag resolution
- Precedence is right: explicit flag > `--from-file` auto-inherit > None. Empty-string / whitespace-only gets `Err("--colab-group cannot be empty")`. `empty_colab_group_errors`, `explicit_colab_group_passes_through_trimmed`, `no_flag_and_no_from_file_returns_none` cover the three branches.
- The comment on the "user types `twapp work` directly (no --from-file)" path is exactly the right design call: ad-hoc direct invocations stay ungrouped, so a human running `twapp work --name audit-helper` from a coordinator's worktree doesn't accidentally inherit the coordinator's group. That would've been a subtle bug.
- Trimming is applied consistently on explicit values but not on inherited parent values (parent's `colab_group: "   "` is treated as unset via the `.filter(|s| !s.trim().is_empty())` clause). Consistent — `from_file_without_parent_colab_group_leaves_field_unset` case-3 covers it.

### `find_spawning_session_colab_group` — upward walk
- Climbs from `cwd` inclusive, stopping at the first `.twapp-session.json`. Returns `None` rather than erroring on read / parse failure, so a corrupted sibling file on the walk doesn't blow up a valid `twapp work` invocation. Right disposition: auto-inheritance is a convenience, not a hard constraint.
- The loop terminates on `parent() == current` (root) — no risk of infinite loop on `/` or a weird mount.
- Reads the JSON with `serde_json::Value::get("colab_group").and_then(as_str)` rather than full `SessionData` deserialization. Good — means a future `SessionData` field change in an unrelated path doesn't break auto-inheritance for existing parent sessions.

### `cmd_claim` — overwrite-already-coordinator flow
- The branching got a little involved: the original "already-coordinator → early return" was inlined. Now it only early-returns when `already_coordinator && colab_group.is_none()` — which means a user can `twapp coordinator claim --colab-group new-group` on an already-coordinator session to update the group without `--force`. `coordinator_claim_updates_colab_group_on_existing_coordinator` pins this. Smart UX — otherwise correcting a group name would require `--force` or a manual JSON edit.
- The output line now differentiates ("Claimed…" vs "Updated… (colab_group=…)") so the operator sees which branch they hit. Good observability.

### `cmd_sessions` column & `format_colab_cell`
- Truncation to 13 chars keeps the row width from exploding on long group names; `colab_cell_long_group_truncated` locks the budget. `colab_cell_none_shown_as_dash` covers unset, empty, and whitespace-only. Good coverage.
- New column widens the divider row from 116 to 137 chars (`"-".repeat(137)`). Matches the added `{:<20}` for the colab cell + 1 space. Arithmetic checks out.

### Fork-path inheritance
- `read_fork_inherited_metadata` now returns a 3-tuple and the fork paths thread `colab_group` through. The existing `cmd_resume` path reads it from `session_data` when writing the forked session. Lock-step with the spec — forks don't spontaneously change group membership.

### GUI plumbing
- `LauncherSession` gains `colab_group` from the real field (not the stub PR #52 kept). These two PRs merge commutatively: whichever lands first, the other's integration becomes a one-line field flip.

## Scope discipline
Files match the briefing: `src-tauri/src/cli/{coordinator,mod,session,session_attribution}.rs` + `gui/sessions.rs` + `README.md` + two skill files. All changes are on the `colab_group` axis; nothing drifting into ui-session-groups or launcher-button territory.

## Tests
All 8 briefing-required tests landed with the exact names listed, plus 8 supporting tests that cover the edge cases that matter (whitespace-only parent value, claim-updates-already-coordinator, column truncation, empty-string rejection on both `--colab-group` code paths). `cargo test --lib` 127/127. Good.

The one-pattern nit: `work_colab_group_flag_sets_field` doesn't actually exercise `cmd_work` or even `create_session_core` — it round-trips a hand-built `SessionData` through `write_session` / `read_session`. The commit explicitly calls this out in a comment ("hermetic … doesn't touch global config / HOME"). That's a reasonable trade — end-to-end would require stubbing HOME and work_directory, which is a lot of apparatus for a single assertion. The `_touch_create_session_core` shim that silences the unused import is worth a follow-up cleanup; either wire at least one real call or drop the import.

## Docs consistency
- README "Co-lab groups" subsection is in the right place (after Roles / provenance), gives the three idiomatic invocations (coordinator launch default, worker auto-inherit, claim override), and the key contracts (empty rejection, direct `twapp work` doesn't auto-inherit). The warning line "legacy session files without `colab_group` load unchanged" sets the right expectation for operators rolling this out mid-stream.
- Reference table updates reflect the new flags on both `coordinator launch` and `coordinator claim`. Good.
- `skills/spawn-agent/SKILL.md` bullet on `--colab-group` is concise and names the auto-inheritance contract. Good.
- `skills/agent-coordinator/SKILL.md` §1b gets the example of `--colab-group infra-refresh` overriding the `--name`-default. Correctly notes "the same coordinator may run multiple disjoint co-labs back-to-back" — that's the one case where the default-to-name heuristic is wrong.

- `find_spawning_session_colab_group` walks `cwd` unconditionally. That means if a user runs `twapp work --from-file brief.md` from inside a nested worktree that happens to have a stale `.twapp-session.json` at a grandparent level, they'd inherit that group. Plausible footgun. Consider limiting the walk depth (e.g. 5 levels) or matching only the direct parent dir's session file. Minor.
- `cmd_claim` prints `"(colab_group=-)"` when `colab_group.is_none()` in the already-coordinator branch — but that branch now short-circuits to the "Nothing to do" early-return before ever hitting the update message, so the `-` fallback is effectively dead. You can drop the `.unwrap_or("-")` without changing behavior.
- The `work_colab_group_flag_sets_field` test comment references "launch_writes_role_coordinator" as the prior art — worth grepping to confirm that test still exists under that name, otherwise the comment points at a ghost.
- Two coordinators running back-to-back with the same `--name` (unlikely but not prevented) would share a `colab_group` with mixed workers. Consider documenting the uniqueness expectation in the skill.

Clean PR. The schema + CLI + inheritance + display add up to the "group membership is tracked as one field" promise; the auto-inheritance heuristic has the right bail-out behavior; PR #52 lines up to consume it cleanly.

PR #57, opened 2026-04-21
feat(ui): urgent message panel + priority filter chip

piekstra:
## Summary

Surfaces the PR #53 priority lane directly in the session window so
urgent/blocker messages don't require a terminal drop-through.

- **`UrgentInbox` panel** (sidebar, above Notes) — polls
  `twapp msg fetch --for <self> --priority urgent|blocker --format json`
  every 10s via a new `fetch_messages` Tauri command. Merges both lanes
  by id (keeps the strongest priority when the symlink surfaces both),
  sorts blocker-first then newest-first.
- **Visual treatment** — red accent on the panel chrome, stronger for
  blockers than urgents. Rows show `from`, subject-or-first-body-line,
  priority chip, and relative time. Clicking a row opens a read-only
  message-view modal.
- **Empty state** — `"No urgent messages."`, auto-collapses after 60s of
  sustained emptiness. User can manually re-expand; polling continues
  regardless so the collapse decision stays fresh.
- **Gating** — panel renders only when the session has a non-default
  handle, so single-session users with no mailbox presence see nothing.

## Out of scope

- **Priority filter chip on the general inbox** — no general inbox view
  exists in the session window yet (only Notes / Prompts / Ticket /
  Urgent), so per the briefing fallback clause this PR is scoped to the
  urgent panel only. The chip row lands with the inbox view.
- **`twapp msg ack`** — the briefing lists an ack button as nice-to-have
  and the CLI doesn't expose an `ack` subcommand today. Deferred.

## Test plan

- [x] `npm test` — 118/118 pass (15 new in `UrgentInbox.test.ts`)
- [x] `cargo test` — 177/177 pass (9 new in `gui::msg::tests` covering
  `build_fetch_argv` + `parse_fetch_stdout` + fake-CLI smoke)
- [x] `npm run build` — clean tsc + vite build
- [ ] Manual (reviewer): spawn a session with a handle, have the
  coordinator send a `--priority blocker` message, confirm the panel
  lights up on next 10s tick. Unit coverage replaces a React Testing
  Library harness (not installed in this repo — matches the
  MessageComposer PR #50 convention of pure-logic vitest tests).

Depends on PR #53 (priority-lane CLI) and PR #50 (composer — Tauri
boilerplate + CSS variable tokens reused here).

### DIFF
--- a/README.md
+++ b/README.md
@@ -642,6 +642,14 @@ trace, not a crash.
 > `twapp msg fetch --priority blocker` is the idiomatic first call at the
 > top of any long write cycle — before the first line of code.
 
+Inside the session window, the sidebar shows an **Urgent** panel directly
+above Notes whenever the session has a handle. It polls `twapp msg fetch
+--for <self> --priority urgent|blocker` every 10s, renders each message
+as a row (from + subject + priority chip + relative time), and auto-
+collapses after the queue has been empty for a minute. Click a row to
+open a read-only message view. Blockers get the strongest red accent,
+urgents a muted one. Single-session users with no handle see no panel.
+
 #### Reading legacy (bare) files
 
 `fetch` accepts both the new fenced-frontmatter shape and the older bare
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -236,6 +236,15 @@ blocker` is exact.
 > `--priority urgent` (or `blocker`) instead; it's the only channel
 > `twapp msg fetch --priority` knows how to find.
 
+When the recipient is running a twapp session window (not a headless
+loop), the priority lane also surfaces visually: the session sidebar
+renders an **Urgent** panel above Notes that polls `--priority urgent`
+and `--priority blocker` every 10s and renders each message as a
+clickable row with the priority chip and relative time. Blockers get the
+strongest red accent. Sending `--priority blocker` is therefore both a
+poll-time hint and an interactive interrupt — the recipient sees the
+panel light up on next poll even without reopening a terminal.
+
 ### Example message
 
 ```
--- a/src-tauri/src/gui/mod.rs
+++ b/src-tauri/src/gui/mod.rs
@@ -114,6 +114,7 @@ pub fn run(args: GuiArgs) {
             monitor::list_monitor_logs,
             files::reveal_in_finder,
             msg::send_message,
+            msg::fetch_messages,
             coordinator::launch_coordinator,
             coordinator::claim_coordinator,
             coordinator::list_claimable_sessions,
--- a/src-tauri/src/gui/msg.rs
+++ b/src-tauri/src/gui/msg.rs
@@ -1,11 +1,15 @@
-//! Tauri command that shells out to `twapp msg send` / `twapp msg broadcast`.
+//! Tauri commands that shell out to `twapp msg`:
 //!
-//! Invoked by the message composer modal (`src/components/MessageComposer.tsx`).
-//! Writes the message body through stdin so no shell-escaping is needed and
-//! large bodies do not hit argv length limits.
+//! - `send_message` — `msg send` / `msg broadcast`. Used by the composer modal
+//!   (`src/components/MessageComposer.tsx`). Writes the body through stdin so
+//!   no shell-escaping is needed and large bodies do not hit argv limits.
+//! - `fetch_messages` — `msg fetch --format json`. Used by the urgent-inbox
+//!   panel (`src/components/UrgentInbox.tsx`), which calls it once per
+//!   priority (urgent + blocker) and merges the results.
 
 use super::types::*;
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
 use std::io::Write;
 use std::process::{Command, Stdio};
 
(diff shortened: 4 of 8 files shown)


piekstra (commented):
**Ship it (COMMENTED / second-pair-of-eyes from `twapp-reviewer`).** CI was still running when I scanned; assuming it lands green this is ready for merge after the manual smoke.

## Policy scan
Clean. No private-project terms. "priority", "blocker", "urgent" are PR-53 vocabulary.

## Correctness

### Backend (`gui/msg.rs::fetch_messages`)
- `build_fetch_argv` rejects unknown priority values (`fetch_argv_rejects_unknown_priority`), lowercases input (`fetch_argv_priority_normalized_to_lowercase`), trims blank `for_handle` (`fetch_argv_skips_blank_for_handle`), and always writes `--format json` — consumer code parses JSON exclusively. Good.
- `parse_fetch_stdout` returns `Ok(Vec::new())` on empty stdout, `Err` on non-array JSON, and `filter_map` drops malformed entries instead of failing the whole batch (`parse_stdout_skips_malformed_entries`). Right disposition for a UI-polling endpoint — a single corrupt message shouldn't blank the whole panel.
- `message_from_value` pulls from the flattened `ParsedMessage` shape (PR-1's `cli/msg.rs`). Required fields `id` / `from` / `ts` use `?` for None-propagation → filtered out by `filter_map`; optional fields default cleanly (`priority` → `"routine"`, `body` / `path` → `""`). Matches the CLI's output contract.
- `FetchedMessage` is `Serialize`-only (no need to round-trip back through the CLI), and uses `skip_serializing_if` on `cc` / `subject` / `thread` so the TS side receives `null` vs. empty distinctions consistently.
- Command runs synchronously on the Tauri command thread — `msg fetch` is I/O-bound on a small directory scan, fast enough. If the mailbox ever grows past a few thousand messages, consider off-thread via `tokio::task::spawn_blocking`, but not today.

### Frontend (`UrgentInbox.tsx`)
- `mergeByPriority` is the key invariant: the `inbox/urgent/<handle>/` symlink lane from PR-53 surfaces blockers under BOTH `--priority urgent` (which returns urgent+blocker per PR-53's semantics) AND `--priority blocker`. The de-dup by id with stronger-priority-wins is exactly right — `dedupes by id, preferring the stronger priority` locks the invariant in tests. Without this, a blocker would appear twice in the panel.
- `priorityRank` returns `routine = 2` for unknown strings, so a future priority token (say `"emergency"`) would show up in the panel but at the bottom of the list rather than crashing. Forward-compatible.
- `parseTs` regex is anchored at the start but not the end (`/^(\d{4})(\d{2})(\d{2})T(\d{2})(\d{2})(\d{2})/`), so a CLI ts like `20260421T090400.123Z` or `20260421T090400Z` both parse. Good — matches PR-1's fractional-second variant mentioned in the ParsedMessage comment.
- `relativeTime` clamps negative deltas to `0s` (future-dated messages don't render "-3s") and passes through unparseable timestamps verbatim. `clamps negative deltas to 0s` + `passes through unparseable timestamps verbatim` both tested. Good.
- `rowPreview`: subject → first non-blank body line → `(no subject)`, truncated to 120 chars. `truncates long previews` test pins the 120 budget.
- `fetchSeqRef` de-dup: every `doFetch` bumps the seq; stale resolves (`if (seq !== fetchSeqRef.current) return;`) are ignored. This is the correct fix for the "two in-flight polls, old one wins" race that React unit-test suites tend to overlook. Nice.
- Polling strategy: runs whether expanded or not so the chevron badge stays live and auto-collapse decisions don't need a new timer when the panel re-opens. Subtle but correct.
- Auto-collapse state machine:
  - `userCollapsed` set on manual collapse → timer never fires until user manually expands again.
  - `emptySinceRef` reset on non-empty, set on transition to empty, timer set for `collapseAfterEmptyMs - elapsed` so re-render doesn't reset the countdown. That's the right shape.
  - One subtle edge case: if `collapseAfterEmptyMs` is changed at runtime (shouldn't happen but props are mutable), the timer uses the new value immediately. Fine.
- `if (!selfHandle) return null;` guard: single-session users without a handle see no panel. `App.tsx` gates with `appConfig?.name && appConfig.name !== "twapp"` so the default-named session doesn't trip the panel either. Layered correctly.
- Message-view modal: read-only, uses `<pre>` for body preservation, backdrop click dismisses, ESC handler — wait, actually I don't see an ESC handler for the viewer. Minor UX — the composer PR #50 had one but this one drops it. Low-cost add.

### CSS
- Three tone classes (`urgent-tone-blocker` / `urgent-tone-urgent` / `urgent-tone-empty`) correctly reflect the strongest priority currently in the panel. Red intensity escalates: blocker uses `0.10` + `0.55` borders, urgent `0.05` + `0.30`, empty has no tint. That's the tuning the briefing asked for.
- `priority-chip` styles have good contrast: `priority-blocker` is solid red on white (`color: #fff`), `priority-urgent` is red text on a 15% red background, `priority-routine` is the muted fallback. Accessible.

## Scope discipline
Files match the PR body exactly:
- Backend: `src-tauri/src/gui/{mod,msg}.rs` (+1, +246/-5)
- Frontend: `src/components/UrgentInbox.{tsx,test.ts}` (+320, +114), `src/App.{tsx,css}` (+4, +259)
- Docs: `README.md` (+8), `skills/agent-coordinator/SKILL.md` (+9)

Per PR body: general-inbox filter chip is correctly deferred (no general inbox view exists), and `twapp msg ack` is deferred (CLI surface doesn't exist yet). Right scope calls.

## Tests
- Rust: 9 new (`fetch_argv_*` x4, `parse_stdout_*` x4, `fetch_fake_cli_success`). 177/177 pass per PR body. Good coverage of argv build + JSON parse + fake-CLI smoke.
- TS: 15 new across `priorityRank`, `mergeByPriority`, `parseTs`, `relativeTime`, `rowPreview`. 118/118 pass per PR body. Pure logic — matches PR #50's convention of unit-testable extracted helpers.

Gaps (not blockers):
- No test for `doFetch`'s two-parallel-invoke behavior — you could inject `fetcher` with a tracker + assert it was called exactly twice per poll (`urgent` + `blocker`). Would catch someone later dropping one branch.
- No test for the auto-collapse state machine. The logic is subtle enough that a one-off vitest with fake timers would earn its keep.
- No test for the `fetchSeqRef` stale-resolver drop — tricky to write without a fake in-flight resolver, but the race is exactly the kind of thing that'd silently regress.

## Docs consistency
- README "Messaging between sessions" section gains the Urgent panel paragraph in the right place (right after the priority-lane fetch examples). Good.
- Skill §3 "Priority lanes" gains a paragraph explaining that `--priority blocker` is now both a poll-time hint AND an interactive interrupt. Correct — this changes the operator mental model in a way that deserves explicit text.
- `--help` output for `twapp msg fetch` is unchanged, correct (no new flags).

- `parse_fetch_stdout` returns `Err` on non-array JSON but the UI currently surfaces this as a red error banner. If the CLI ever produces a plain error-string on some failure mode, the panel would look "stuck in error". Consider catching a well-known error-object shape from the CLI and mapping it to a friendly message — or leave as-is, the fake-CLI test proves the happy path.
- The viewer modal has no `ESC-to-close` keyboard handler. `onClick={() => setViewing(null)}` on the backdrop + explicit `x` button are both there, but mac-native modal-close convention is ESC.
- `parseTs` regex allows fractional seconds positionally but the capture groups only include integer seconds; a ts with `.123Z` would parse to the second boundary, losing millisecond precision in the relative-time display. Not user-visible (1ms in a `1s` display is invisible).
- `App.tsx` gates on `appConfig.name !== "twapp"`. If the default session name ever changes (say to `twapp-default`), this string comparison silently re-enables the panel for single-session users. Worth a constant `DEFAULT_SESSION_NAME` or an explicit `appConfig.has_handle: bool` field.
- The `section-refresh-btn` in the header is a generic class from elsewhere — fine, but the unicode `⟳` (`&#8635;`) is not announced well by screen readers. An `aria-label="Refresh urgent queue"` is already on the `title` attr; consider a visible `sr-only` text too.

Solid PR. The merge-by-priority invariant is the one nontrivial bit and it's well-tested; the auto-collapse state machine is the riskiest bit and it has two gotchas guarded (userCollapsed vs auto, emptySinceRef reset semantics). Ready to merge after the reviewer's manual smoke — would especially like to see a coordinator-sends-blocker → panel-lights-up within one 10s tick confirmation.

piekstra (commented):
**Ship it.** Right scope per briefing (urgent-panel only; priority filter chip deferred to a later inbox-view PR), dense pure-helper test coverage, single-session preservation honored via `selfHandle` gate.

## Policy scan

Clean.

## What's right

### Backend (`gui/msg.rs`)

- **`fetch_messages` extends the `gui/msg.rs` file** instead of creating a new module. Sensible grouping — send and fetch share `twapp_binary()`, the shell-out plumbing, and the fake-CLI test pattern. The module docstring was updated to describe both commands.
- **`build_fetch_argv` is a pure function** with 5 argv tests: minimal, all-fields-present, lowercase normalization, unknown-priority rejection, blank-for_handle skipping.
- **`parse_fetch_stdout` robustness**:
  - Empty / whitespace-only stdout → `Ok(vec![])` (right call — `twapp msg fetch --priority urgent` on an empty urgent queue is the common success path).
  - Non-array JSON → typed error with the received value in the message.
  - Malformed entries skipped rather than failing the batch — one missing `id` doesn't wipe out a live blocker in the same response.
- **`message_from_value`** parses the flat `ParsedMessage` shape that `cli::msg` produces directly, with defensible defaults (priority defaults to "routine", body/path default to empty string). ✓

### Frontend (`UrgentInbox.tsx`)

- **Pure helpers exported for testing**:
  - `priorityRank` — blocker < urgent < routine; unknown treated as routine. Tested.
  - `mergeByPriority` — deduplicates by id, preserving strongest priority when the urgent-lane symlink surfaces the same message under both `priority=urgent` and `priority=blocker`. That's the messaging-design priority-lane hardlink pattern; the merge correctly handles it. Sort is blocker-first then newest-first; `string.localeCompare` on `YYYYMMDDTHHMMSSZ` timestamps matches chronological order (no dashes, no timezone — pure lex compare works).
  - `parseTs` — parses the CLI's compact `YYYYMMDDTHHMMSSZ` format directly (since that string isn't a valid ISO 8601 for `Date.parse`). Returns `null` on malformed.
  - `relativeTime` — formats seconds/minutes/hours/days with a negative-delta clamp so future-dated messages render as `0s` rather than nonsense. Clock injected for deterministic tests.
  - `rowPreview` — subject first, else first non-empty body line truncated at 120 chars, else `(no subject)`.
- **`fetchSeqRef` dedupes stale in-flight fetches** — common async-react pattern; prevents a slow response from overwriting a fresh one.
- **`useEffect` auto-collapse** correctly separates `userCollapsed` from "collapsed because empty for 60s":
  - User-collapse sets the flag and kills the auto-collapse timer.
  - Manual re-expand clears the flag and restarts the emptySince timer.
  - Good separation: respects user intent, doesn't fight them.
- **Polling runs whether expanded or not** so the chevron count stays live even when collapsed. Documented in a comment.
- **Fetcher + clock injected as props with sensible Tauri/window defaults** — the test file can substitute both without monkey-patching globals. That's the right test seam.

### Gating

- **`selfHandle: null` → no panel**. No mailbox presence, no urgent UI. The agent-aware-ui design §3.6 single-session preservation rule is honored at the component boundary, not buried inside render logic. ✓

### Test coverage

- 5 describe-blocks cover every pure helper: `priorityRank` (ordering + unknown), `mergeByPriority` (3-message order, dedup-strong-priority, both-empty, single-empty-alongside-populated), `parseTs` (format + malformed), `relativeTime` (s/m/h/d + negative clamp + unparseable passthrough), `rowPreview` (subject / body-first-line / empty).
- Rust: 10 new tests covering argv shape, priority normalization, validation, stdout parsing (empty / happy / non-array / malformed-skip), plus a fake-CLI integration test.

## Scope discipline

- PR body explicitly names the priority filter chip as out-of-scope ("no general inbox view exists in the session window yet") and `twapp msg ack` button as out-of-scope (briefing nice-to-have, CLI not yet exposing the verb).
- README and `agent-coordinator/SKILL.md` updates are small (+8 and +9 lines) — they document the new panel without expanding its surface.

## Status

- CI: `SUCCESS` (completed 09:19:20Z).
- `mergeStateStatus: UNSTABLE` only because no approving review — will flip on merge.

## Non-blocking nits

- **`UrgentInbox` polling** runs whether the user can see the panel or not (the component renders even when `selfHandle` is null — it just early-returns in the effect). An earlier return before `useEffect` registration would halve the number of noop polls for single-session users. Non-blocking; today's 10s poll with a null-guard is cheap.
- **`mergeByPriority` tiebreak** uses `ts.localeCompare` — works because `YYYYMMDDTHHMMSSZ` is a pure lex-orderable format. If the CLI ever emits an alternate timestamp shape (e.g., sub-second precision with dots), the comparator would still work but the test assertions would need updating. Worth a one-line comment pointing at `cli::msg`'s timestamp emitter as the source of truth.
- **`relativeTime` caps at days** — if you ever surface messages older than a month (`--limit` is typically small), `>=30d` would render as e.g. `45d`. Acceptable; urgent queue by nature is short-lived.
- **Row click → read-only message-view modal** — couldn't see the modal code in the slice I read, but the PR body describes it. Worth making sure it doesn't pull the whole message body into the DOM for 100+-line bodies; truncate + "expand" if needed. Non-blocking for today's use.

## Summary

Urgent panel wires a real Tauri command shell-out + rich pure-helper testing; selfHandle gating preserves the single-session lean UI; test coverage pins the merge + sort contracts so #53's priority-lane-symlink behavior stays visible. Ship it.

PR #58, opened 2026-04-21
feat(msg): threading (reply-to + twapp msg thread) — design PR-2

piekstra:
## Summary

- `twapp msg thread <thread-id>` lists every message in the thread in chronological order (both messages whose `thread:` equals `<id>` and the root whose `id` equals `<id>`). `--format json` supported.
- `twapp msg send --reply-to <parent-id>` already lands in PR-1: inherits `thread` from parent (or uses parent's own id if the parent is a root), sets `in_reply_to: <parent-id>`, errors if parent is missing. This PR exercises it with the named tests.
- Design: [`docs/designs/agent-messaging.md`](../blob/main/docs/designs/agent-messaging.md) §2.4, §5 PR-2.

## Scope

In:
- New `MsgCommands::Thread { thread_id, format }` subcommand + `cmd_thread` / `list_thread` impl.
- Five PR-2 tests from the design:
  - `reply_to_inherits_thread_from_parent` — multi-hop: leaf inherits root thread via an intermediate reply (not the intermediate's id).
  - `reply_to_new_thread_uses_parent_id_as_root` — parent is root, reply's thread == parent's id.
  - `reply_to_nonexistent_parent_errors` — error mentions the id + "not found".
  - `thread_lists_in_chronological_order` — unrelated messages excluded; timestamps non-decreasing.
  - `thread_returns_empty_on_unknown_id_without_crash` — also tolerates an empty inbox.

Out (design-doc PR-3+):
- Directory split (`inbox/direct/`, `inbox/broadcast/`) — PR-3.
- Priority lane filter work — PR-4.
- Optional `threads/<thread-id>/<ts>-<id6>.md` symlink secondary index — skipped as nice-to-have to keep PR small; can follow up if useful.

## Test plan

- [x] `cargo test --lib cli::msg::` → 38 passed, 0 failed, including all 5 new tests.
- [x] `cargo build --bin twapp` clean.
- [x] `twapp msg thread --help` renders the new subcommand.

### DIFF
--- a/README.md
+++ b/README.md
@@ -580,8 +580,14 @@ twapp msg send reviewer "PR-1 is up, ready to review"
 twapp msg send worker-a,worker-b --priority urgent --subject "build broke" "see CI 1234"
 twapp msg send reviewer --cc coordinator,qa "heads up on scope change"
 
-# Replies inherit the parent's thread id and set in_reply_to.
+# Replies inherit the parent's thread id and set in_reply_to. If the
+# parent is itself a root (no thread: field), the reply's thread id
+# becomes the parent's own id. Replies at any depth thread to the root.
 twapp msg send reviewer --reply-to 01JS4M7Q8W "ack, rebasing now"
+
+# List every message in a thread (root + all replies), chronologically.
+twapp msg thread 01JS4M7Q8W
+twapp msg thread 01JS4M7Q8W --format json | jq
 
 # Broadcast — writes to: [all], or to: [channel:<name>] with --channel.
 twapp msg broadcast "standup in 5"
@@ -922,6 +928,7 @@ burns iteration.
 | `twapp msg send <to> [body]` | Send a message (writes to the shared mailbox inbox) |
 | `twapp msg broadcast [body]` | Broadcast to every handle (`to: [all]`) |
 | `twapp msg fetch [--for <h>] [--since <ts>]` | List inbox messages, filtered |
+| `twapp msg thread <thread-id>` | List every message in a thread chronologically; `--format json` for machine-readable |
 | `twapp msg claim <lane-id> [--note <s>]` | Atomically claim a shared lane (PR, audit, backlog item); exit 1 if already claimed |
 | `twapp msg release <lane-id> [--note <s>]` | Release a lane you own; writes `released.json` and broadcasts the release |
 | `twapp msg claim --list [--lane-prefix <p>]` | List all active (unreleased, unstale) claims; `--format json` for machine-readable |
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -245,6 +245,30 @@ strongest red accent. Sending `--priority blocker` is therefore both a
 poll-time hint and an interactive interrupt — the recipient sees the
 panel light up on next poll even without reopening a terminal.
 
+### Threading (`--reply-to` + `msg thread`)
+
+Long coordinator ↔ worker conversations span multiple messages. Keep
+them coherent by threading instead of leaning on `re:` prose:
+
+```bash
+# Start a thread (the root is whatever you sent first — no flag needed).
+twapp msg send worker-a --subject "scope change" "<body>"
+
+# Reply to a specific parent — inherits the parent's thread id and sets
+# in_reply_to. If the parent is itself a root (no thread: field), the
+# reply's thread id becomes the parent's own id.
+twapp msg send worker-a --reply-to 01JS4M7Q8W "ack, rebasing now"
+
+# List every message in a thread chronologically (root + all replies).
+twapp msg thread 01JS4M7Q8W
+twapp msg thread 01JS4M7Q8W --format json | jq
+```
+
+Use this for REDIRECTs, scope changes, and multi-turn reviews — anywhere
+the second message only makes sense in the context of the first. The
+thread id is stable for the lifetime of the conversation; a reply at
+depth N still threads to the root, not to the immediate parent.
+
 ### Example message
 
 ```
--- a/src-tauri/src/cli/mod.rs
+++ b/src-tauri/src/cli/mod.rs
@@ -528,6 +528,7 @@ pub fn run(cmd: Commands) -> i32 {
                     msg_archive::cmd_list(since, format)
                 }
             },
+            MsgCommands::Thread { thread_id, format } => msg::cmd_thread(thread_id, format),
         },
         Commands::Models { command } => match command {
             ModelsCommands::List { provider, format } => models::cmd_list(provider, format),
--- a/src-tauri/src/cli/msg.rs
+++ b/src-tauri/src/cli/msg.rs
@@ -214,6 +214,20 @@ pub enum MsgCommands {
         #[command(subcommand)]
         command: super::msg_archive::ArchiveCommands,
     },
+    /// List every message in a thread in chronological order.
+    ///
+    /// Matches both messages whose `thread:` equals `<thread-id>` and the
+    /// root message whose `id:` equals `<thread-id>` (thread roots carry no
+    /// `thread:` field on their own). Prints a short one-line-per-message
+    /// summary by default; `--format json` dumps the full parsed messages.
+    #[command(after_help = "Examples:\n  twapp msg thread 01JS4K2AAA\n  twapp msg thread 01JS4K2AAA --format json")]
+    Thread {
+        /// Thread id (root message id).
+        thread_id: String,
+        /// Output format.
+        #[arg(long, value_enum, default_value_t = FetchFormat::Pretty)]
+        format: FetchFormat,
+    },
 }
 
 // --- Mailbox discovery ------------------------------------------------------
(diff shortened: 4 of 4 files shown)


piekstra (commented):

## Policy scan
Clean. No private-project terms. "thread", "reply", "root" are messaging-design vocabulary.

## Correctness

### `list_thread` — membership rule
- Single retain clause: `m.fm.id == thread_id || m.fm.thread.as_deref() == Some(thread_id)`. Matches design §2.4 exactly: the root carries no `thread:` on itself (its `id:` IS the thread id), and every reply in the tree carries `thread: <root-id>` regardless of nesting depth. This handles both the "root case" and the "intermediate + leaf reply case" without needing a separate code path.
- Preserves the chronological ordering of `list_messages`. `thread_lists_in_chronological_order` asserts `thread[0].fm.ts <= thread[1].fm.ts`, locking the invariant in.

### Reply-to inheritance (exercised via `send_reply` helper)
- `thread_id = parent.thread.clone().unwrap_or(parent.id.clone())` — correctly propagates the ROOT's id through any depth of reply chain rather than using the immediate parent's id. This is the one subtle bit in threading protocols: a naive implementation would use `in_reply_to` as the thread key, which would fork the thread into N sub-trees. `reply_to_inherits_thread_from_parent` pins this — `leaf.fm.thread == root.fm.id`, not `mid.fm.id`. Good coverage.
- `reply_to_new_thread_uses_parent_id_as_root` locks the root-case: parent has no `thread:`, so the reply's thread is parent's id.
- `in_reply_to` vs `thread` distinction is maintained in the reply:
  - `in_reply_to: parent.id` — the immediate predecessor (for reply-to-this-message reasoning)
  - `thread: parent.thread.unwrap_or(parent.id)` — the root-anchored thread key (for list-whole-thread reasoning)
- `reply_to_nonexistent_parent_errors` checks that the error names both the id AND says "not found" — the one-line error shape future callers will grep for.

### CLI surface
- `twapp msg thread <thread-id> [--format json|pretty]`. `--format pretty` (default) prints a header line plus one line per message with ts / id / from / priority / subject. `--format json` dumps the full `Vec<ParsedMessage>`. Good — matches the convention of `msg fetch`.
- `msg thread --help` gets an `after_help` with two examples. Consistent with `msg archive` subcommands.
- Exit codes: `0` for success-or-empty, `1` only for mailbox-resolution or JSON-serialize failures. No `2` (conflict) exit on empty thread — `(no messages in thread {})` is a valid lookup result, not an error. Right call.

### Tests
All 5 briefing-required tests land by exact name:
- `reply_to_inherits_thread_from_parent` ✓ (multi-hop inheritance)
- `reply_to_new_thread_uses_parent_id_as_root` ✓ (root case)
- `reply_to_nonexistent_parent_errors` ✓ (error contains id + "not found")
- `thread_lists_in_chronological_order` ✓ (includes a decoy unrelated message to catch false positives)
- `thread_returns_empty_on_unknown_id_without_crash` ✓ (plus empty-inbox sub-case)

The `send_reply` test helper replicates `cmd_send`'s reply-to path in a hermetic way. A unit-test decision worth calling out: it bypasses the CLI's env/session plumbing so we can test the logic without a fake session dir. The risk here is that the helper and production drift. Consider cross-testing against `cmd_send` via one sentinel test that builds a real `SendArgs` → `cmd_send` call, to confirm production and helper produce the same `thread` / `in_reply_to` frontmatter. Not a blocker for PR-2.

## Scope discipline
Files match the PR body exactly: `src-tauri/src/cli/{mod,msg}.rs`. Two files, +194 lines, zero deletions. Touches nothing in archive/claim/priority-lane territory (PR-3/4/7). The punted `threads/<thread-id>/<ts>-<id6>.md` symlink index is a reasonable deferral — `list_thread` is O(n) over the inbox today but thread queries are rare enough in the current traffic pattern that it won't bite for a while.

## Docs consistency
No doc changes in this PR. That's a **miss** worth flagging — the README's "Messaging between sessions" section currently describes PR-1 + PR-4 only (from #53's update). With this landing, `twapp msg thread` becomes a first-class subcommand that deserves at least a one-line mention in the README + the quick-reference table. Not blocking merge, but please fold a doc update either into this PR or a follow-up — the CLI is the primary operator surface and gaps in README drift into "nobody knows this exists" territory fast.

Minor nit in the same spirit: `skills/agent-coordinator/SKILL.md` §3 ("Messaging") is the natural place for a bullet on `--reply-to` + `msg thread` as the idiomatic way to keep a multi-turn coordinator↔worker conversation coherent. That's a coordinator-workflow change, not just a CLI addition.

- `list_thread` does an O(n) scan of the flat inbox. Once PR-3's directory split lands (`inbox/direct/`, `inbox/broadcast/`), this scan will stay flat against both subdirs — worth revisiting at that point, but fine for PR-2.
- `print_thread_summary` hardcodes the column order. If a future operator wants `--format short` vs `--format pretty`, a small formatter struct would make that easy; overkill today.
- `cmd_thread` doesn't accept `--since` or `--limit`. Design call: threads are typically short enough that full-list is the right default, and a caller who wants a slice can pipe `--format json` into `jq`. Fine.
- The empty-state messages (`"(no messages in thread {})"` and `"(nothing to rotate)"` in msg_archive) use parens+lowercase consistently. If you care about that idiom, consider a shared helper; otherwise the repetition is fine.

Solid, tightly-scoped PR. The one subtle thing (root-anchored thread id vs immediate-parent-id) is well-tested. CI is green, the merge state is CLEAN. Main follow-up ask is the README + skill doc update — every PR-1/4/5/7 in this design landed with docs; #58 shouldn't break that pattern.

piekstra (commented):
**Ship it — docs follow-up is clean.**

## Rescan
- Delta since my previous review (a3635e76 → b507a296): one docs commit touching `README.md` (+8/-1) and `skills/agent-coordinator/SKILL.md` (+24).
- Policy scan: clean. No private-project terms.

## What landed
- **README quick-reference table** gets a `twapp msg thread <thread-id>` row next to the other `msg` entries. Discoverable from the same table operators already scan.
- **README messaging section** gains two code-example additions: reply-to semantics (clarifies the parent-is-root case with the right mental model — "replies at any depth thread to the root"), and `twapp msg thread` usage with pretty + JSON formats.
- **Coordinator skill §3** gains a new "Threading (`--reply-to` + `msg thread`)" subsection with the three-idiom example (start / reply / list) and the operator-facing guidance ("Use this for REDIRECTs, scope changes, and multi-turn reviews — anywhere the second message only makes sense in the context of the first"). That's exactly the workflow hook I'd hoped for.

The "thread id is stable for the lifetime of the conversation; a reply at depth N still threads to the root, not to the immediate parent" sentence is the operator-facing version of the test invariant pinned by `reply_to_inherits_thread_from_parent`. Nicely matched.

Nothing else to flag. Merge when ready.

piekstra (commented):
**Ship it.** Right scope for design PR-2, CI green, tests pin the five named scenarios from the design doc.

## Policy scan

Clean.

## What's right

- **Scope matches design §5 PR-2 exactly**: the `twapp msg thread` subcommand + full test coverage for the `--reply-to` logic (which PR-1 already shipped in main). No scope creep — directory split, priority lane filter, and `threads/<id>/` symlink index are explicitly deferred to PR-3/PR-4/nice-to-have.
- **`list_thread` retention predicate** is the right shape: matches either `m.fm.id == thread_id` (the root case — roots carry id but not `thread`) or `m.fm.thread == Some(thread_id)`. Covers the design §2.4 definition that roots set `thread: <self-id>` implicitly by identity. ✓
- **`cmd_thread` output** has two cleanly-separated formats: a summary table (one-line per message: ts / id / from / priority / subject) and `--format json` for scripting. Empty case prints `(no messages in thread <id>)` to stdout with exit 0 — right call, "unknown thread" isn't an error, it's just a zero-match query.
- **All 5 named PR-2 tests present**:
  1. `reply_to_inherits_thread_from_parent` — **multi-hop thread inheritance**: leaf correctly inherits root's thread id via an intermediate reply, not the intermediate's own id. This is the load-bearing case — depth-N replies thread to the root, not to the immediate parent. The test explicitly asserts `leaf.fm.thread == Some(root.fm.id)` and `leaf.fm.in_reply_to == Some(mid.fm.id)`. Both must be true for threaded UI to render correctly.
  2. `reply_to_new_thread_uses_parent_id_as_root` — parent-is-root case: reply's thread id = parent's own id. Asserts `root.fm.thread.is_none()` first to pin the root contract.
  3. `reply_to_nonexistent_parent_errors` — error names the missing id + says "not found". Both checked explicitly, so a future refactor that drops either half will break the test.
  4. `thread_lists_in_chronological_order` — inserts an unrelated message to verify leakage doesn't happen; timestamps strictly non-decreasing.
  5. `thread_returns_empty_on_unknown_id_without_crash` — plus the empty-inbox edge case in the same test.

## Small observations

- **The test helper `send_reply`** replicates `--reply-to`'s logic (`parent.thread.clone().unwrap_or(parent.id.clone())`) rather than going through `cmd_send` which requires env setup. That's a pragmatic choice given the CLI's env-var mailbox discovery, but it means if `cmd_send`'s reply-to path ever drifts from this 8-line helper, the tests here stay green while production breaks. The drift risk is small (the helper is tiny and colocated), but worth a comment saying "this must mirror `cmd_send`'s reply-to branch" or — better — have `cmd_send` extract the same helper and share it. Non-blocking.
- **1100ms sleeps** between sends to ensure distinct `YYYYMMDDTHHMMSSZ` timestamps at seconds precision. Right workaround for the format's resolution; makes the test slow (~4s total across the chronological-order test) but deterministic. Acceptable.
- **`print_thread_summary` formatting** uses tab-free space separation — `awk`/`cut`-friendly. The columns are aligned visually in the test fixture but not enforced — a very-long subject will push the line wide. Fine for terminal viewing; scripters should use `--format json`.
- **`MsgPriority::Routine` on replies** — `send_reply` hardcodes routine priority. Design doesn't say replies must inherit parent priority; `--priority` on `send` still wins if the reply is urgent on its own. Consistent with §2.4's "threads are independent of priority lanes" framing.

## Skill doc update

- `agent-coordinator/SKILL.md` gains a "Threading (`--reply-to` + `msg thread`)" subsection under §8 with a three-command example. Right placement — the skill already covers `--priority urgent/blocker` escalation next to it, so thread+priority are now adjacent in the operating manual. Framing is load-bearing: "thread id is stable for the lifetime of the conversation; a reply at depth N still threads to the root, not to the immediate parent" — that's exactly the guarantee the multi-hop test pins.

## README

Updated example block shows `--reply-to` inherits thread / falls back to parent id, and adds `twapp msg thread <id>` + `--format json | jq` examples. Consistent with the skill's phrasing. ✓

## Summary

Focused PR-2 implementation, five design-named tests all present, multi-hop inheritance explicitly exercised (the load-bearing case), CLI + skill + README updates consistent. CI green, MERGEABLE.

**Ship it.** Sets up PR-3 (directory split + cursors) with the threading fields production-ready.

PR #60, opened 2026-04-21
feat(msg): directory split + read cursors — design PR-3

piekstra:
## Summary

Implements **PR-3** of [`docs/designs/agent-messaging.md`](../blob/main/docs/designs/agent-messaging.md) (§2.1, §2.7, §2.9, §2.10): per-recipient directory split for `inbox/`, append-only read/ack cursors per handle, and a one-shot `migrate` to move any pre-PR-3 flat files into the new layout. Backwards-compatible via a grace-period symlink shim and dedup-on-read.

- **Directory split** — `inbox/broadcast/`, `inbox/direct/<handle>/`, `inbox/channel/<name>/`. Multi-recipient sends land canonical + per-recipient symlinks; direct `cc:` also gets a `direct/<cc>/` symlink so cc'd workers find it by scoped scan.
- **Legacy shim** — every send also writes `inbox/<filename>.md` as a symlink back to the canonical file during the grace period, so un-upgraded readers doing `ls inbox/` continue to work (design §2.10).
- **Cursors + ack** — `<mailbox>/cursors/<handle>.jsonl` append-only. `fetch --mark-read` records a `read` per returned message; `msg ack <id>` records an `ack`. Default `--since` for a fetch with a handle is strictly-after the last `read` — a worker can \`fetch --mark-read\` in a loop and only see new traffic.
- **Migrate** — \`twapp msg migrate [--dry-run] [--drop-legacy]\` moves every regular `inbox/*.md` into its canonical slot, laying down grace-period symlinks at the old flat path. Idempotent. `--drop-legacy` closes the period.
- **Urgent lane** (PR-4) retargeted at the canonical under the new layout; broken-symlink and multi-recipient tests still pass.

## Test plan
- [x] Subdir-based send creates canonical file + legacy symlink
- [x] Multi-recipient direct creates one file/symlink per recipient
- [x] \`fetch\` sees messages in new shape
- [x] \`fetch\` also sees legacy-flat messages during grace period
- [x] \`fetch --since <cursor>\` narrows correctly (including against legacy flat)
- [x] \`fetch --mark-read\` appends the right cursor entries
- [x] \`ack\` appends \`action:\"ack\"\` (with and without \`--note\`)
- [x] \`migrate\` moves flat files by \`to:\` field; idempotent re-run is a no-op
- [x] \`migrate --drop-legacy\` removes grace-period symlinks
- [x] Default \`fetch --since\` resumes strictly-after last read cursor
- [x] Full suite: \`cargo test --lib\` — 212 passed, 0 failed

## Out of scope
Presence/heartbeat (PR-5), channel subscription model (PR-6). Archive rotation shape (PR-7) is left alone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,15 +548,17 @@ filesystem mailbox is a simple way to let them leave each other messages
 without leaving twapp.
 
 `twapp msg` is a thin wrapper over that convention: it drops
-fenced-frontmatter markdown files into a shared `inbox/` directory and
-reads them back. See
+fenced-frontmatter markdown files into a shared mailbox and reads them
+back. See
 [`docs/designs/agent-messaging.md`](docs/designs/agent-messaging.md) for
-the full model. Today's CLI covers **PR-1** (scaffolding) and **PR-4**
-(priority lane) of that design — `send`, `broadcast`, and `fetch`
-against the current flat `inbox/` layout, plus urgent-lane symlinks
-under `inbox/urgent/<recipient>/` for time-sensitive traffic.
-Threading, directory split, cursors, presence, channels, and archive
-rotation all land in later PRs.
+the full model. Today's CLI covers **PR-1** (scaffolding), **PR-2**
+(threading), **PR-3** (directory split + read cursors), **PR-4**
+(priority lane), and **PR-7** (archive rotation) of that design. Sends
+land under a per-recipient layout (`inbox/broadcast/`,
+`inbox/direct/<handle>/`, `inbox/channel/<name>/`); every send also
+drops a grace-period symlink at the flat `inbox/<filename>.md` path so
+un-upgraded readers doing `ls inbox/` keep working. Presence and
+first-class channel subscriptions still land in later PRs.
 
 #### Configure the mailbox
 
@@ -570,7 +572,22 @@ export TWAPP_MAILBOX_DIR="$HOME/collab/mailbox"
 export TWAPP_SHARED_DIR="$HOME/collab"
 ```
 
-Messages land in `<mailbox>/inbox/<YYYYMMDDTHHMMSSZ>-<id6>.md`.
+Messages land under the per-recipient split layout (design §2.1):
+
+```
+<mailbox>/inbox/
+  broadcast/<ts>-<id6>.md          # to: [all]
+  direct/<handle>/<ts>-<id6>.md    # to: [<handle>]
+  channel/<name>/<ts>-<id6>.md     # to: [channel:<name>]
+  urgent/<handle-or-all>/…         # priority lane (PR-4)
+  <ts>-<id6>.md                    # grace-period symlink → canonical
+```
+
+A send always writes one canonical file and leaves symlinks for every
+additional recipient, every direct `cc:`, and the legacy flat path.
+`twapp msg migrate` (below) moves any pre-PR-3 flat files into the new
+slots; `migrate --drop-legacy` closes the grace period by removing the
+flat symlinks once every participant is on the new layout.
 
 #### Send, broadcast, fetch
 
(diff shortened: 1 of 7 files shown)


piekstra (commented):
**Ship it.** Implements design §2.1 + §2.7 + §2.9 + §2.10 in one PR; grace-period symlink shim keeps un-upgraded readers working; cursor semantics (`read` vs `ack`, `--mark-read` opt-in, default strict-`>` since) match the design precisely.

## Policy scan

Clean on the extended regex.

## What's right

### Directory split (§2.1)

- **`recipient_path`** routes by prefix:
  - `"all"` → `inbox/broadcast/<filename>`
  - `"channel:<name>"` → `inbox/channel/<name>/<filename>`
  - any other handle → `inbox/direct/<handle>/<filename>`
  - Empty or malformed (`channel:` with empty name) → `None`
  The explicit prefix-match is the right shape — no regex, no string guessing, no ambiguity on handles that happen to contain the substring "channel".
- **Canonical-file-plus-symlinks**: `write_message` lands the real file at the first routable recipient's slot, then symlinks for every additional `to:` and every direct `cc:` (channel and `all` cc skipped — correct per design §2.2). Multi-recipient traffic lives once on disk.
- **Legacy symlink at `inbox/<filename>`** automatic for every send. Grace-period readers doing `ls inbox/` see everything. `migrate --drop-legacy` closes the period.
- **`symlink_to_canonical`** builds a *relative* symlink via `make_relative` so moving the mailbox tree doesn't break links. Idempotent (existing file/symlink → no-op). Best-effort — failures are `log::debug!`d rather than bubbled up, matching the existing "symlinks are convenience, filesystem state is authoritative" posture.
- **`list_messages` dedup**: `HashSet<PathBuf>` of canonical paths keeps broadcast+direct multi-recipient sends + legacy symlinks from double-counting. `std::fs::canonicalize` on each entry ensures symlink targets collapse to the same seen-key as the canonical file. ✓
- **The `legacy_symlink_is_not_double_counted_even_when_dangling` test** locks in the right behavior: a dangling symlink at `inbox/X.md` + the canonical file at `inbox/direct/X/X.md` produces exactly one `ParsedMessage`, not two.

### Cursors (§2.7)

- **`cursors/<handle>.jsonl` append-only**, one JSON object per line. Each entry's `ts` is the *message's* frontmatter timestamp, not the action wall-clock — so it's a drop-in `--since` value (documented in the doc string + skill + README). That's the key design decision from §2.7 and it's honored here.
- **`cmd_ack`** is a distinct subcommand, not a flag on `fetch`. Matches the design's intent that `ack` is a commitment, not a side effect of reading.
- **`--mark-read` is opt-in**: `fetch` without the flag is a peek. Test `explicit_since_stays_inclusive_and_does_not_consume_cursor` verifies the cursor file isn't created by a plain fetch even with `--since` set.
- **Default `--since` is strictly-`>`**: a handle with a previous `read` cursor fetches only strictly-newer messages on the next `fetch`. Explicit `--since X` is inclusive (≥), so passing a known cursor ts re-surfaces its message (crash recovery). The `fetch_default_since_picks_up_where_last_read_left_off` test pins both sides: default skips already-read, explicit re-surfaces.

### Migrate (§2.10)

- **`cmd_migrate`** is **idempotent** — re-running finds nothing to move. Test coverage explicit.
- **`migrate_skips_file_without_to_field`** — defensive; broken files don't wedge the migration.
- **`migrate_bare_legacy_file_routes_by_header_to`** — picks up the permissive parser from PR-1 and routes bare legacy files (the 84% of pre-PR-1 corpus shape) without requiring manual cleanup first.
- **`--drop-legacy`** removes every symlink directly under `inbox/`. Separate flag from the move step so operators can verify the split layout looks right before committing.

### Backward compat

- The dedup-on-read in `list_messages` is the whole reason the grace-period symlinks work — a reader that hasn't upgraded still `ls inbox/`s and finds every message (via the legacy symlinks), and a reader that has upgraded doesn't double-count. The test `list_messages_sees_split_layout_and_flat_legacy_without_duplicates` explicitly stages both shapes in the same mailbox and asserts on the deduped result.
- The `fetch_since_on_legacy_flat_message_filters_correctly` test ensures legacy-flat messages coming through the fallback path still honor `--since` — important because a migration hasn't happened yet means old content is still flat.
- **PR-4 urgent lane** and **PR-1 msg_claim broadcast shadow** updates: `create_urgent_symlinks` now targets the canonical under the new layout, and `msg_claim` tests assert `inbox/broadcast/<filename>.md` exists instead of `inbox/<filename>.md`. Both downstream surfaces audited.

### Test surface

- 18+ new tests in `cli::msg::tests::*` covering: direct/broadcast/channel canonical paths, legacy symlink creation, CC symlinks, dedup on read, dangling legacy symlinks, mark-read cursor writes, default-since exclusive skip, explicit-since inclusive, cursor-file-not-created-without-mark-read.
- 8 migrate tests: flat→direct, broadcast, channel, multi-recipient, idempotent, drop-legacy, dry-run, skip-missing-`to`, bare-legacy routing.
- `msg_claim` tests updated to read from `inbox/broadcast/` rather than flat `inbox/` — downstream consistency verified.

## Non-blocking nits

- **`make_relative`** returns `None` only when one path is absolute and the other is relative with no overlap. If both are absolute but on different drives (not really a thing on POSIX), it'd still produce a useless relative path. Fine in practice; worth a one-line comment.
- **`symlink_to_canonical`** uses `std::os::unix::fs::symlink`. Windows won't compile. twapp is darwin-first (and Linux-compatible), so acceptable. Worth a `#[cfg(unix)]` on the function if someone ever wants cross-platform compilation later.
- **`scan_flat_dir_into` calls `std::fs::canonicalize` on every entry** — that's one syscall per file during scan. For a small mailbox (<100 messages) it's fine; for a growing archive, this could become a hot path. Profile and move to an mtime-based pre-filter if it ever shows up. Non-blocking today.
- **`list_messages` sort by `(ts, id)`** preserves determinism. Good — the test suite relies on this for `chronological_order` assertions.
- **The 1100ms sleep** between sequential `write_message` calls in tests (to make `YYYYMMDDTHHMMSSZ` timestamps strictly increase) slows this test file considerably. Acceptable for correctness, but worth noting — if anyone adds 20 more tests with the same pattern, the `cargo test` duration starts to bite.
- **Cursor file size**: `cursors/<handle>.jsonl` grows without bound. Design §2.7 says rotation is "same as archive" (§2.8). PR-7 handled archive rotation; cursor rotation is implied but not explicitly part of this PR. Worth a future follow-up to prune cursor entries older than the archive retention.

## Scope discipline

- PR body names what's in (directory split + cursors + ack + migrate) and what's deferred (presence and first-class channel subscriptions still land in later PRs). The urgent-lane retarget is noted as PR-4 behavior preservation, not a new feature.
- README + skill updates track the new layout accurately. The SKILL's `## Archiving` section updates `mv <inbox>/<msg>` → `mv <inbox>/direct/<me>/<msg>` to match the canonical path.

## Status

- CI: SUCCESS.
- `mergeStateStatus: CLEAN`.

## Summary

Big, well-tested refactor that advances the messaging substrate to design §2.1's per-recipient layout while keeping pre-PR-3 readers functional via the legacy symlink shim. Cursor semantics match §2.7 (read vs ack, opt-in mark-read, strict-`>` default-since). `migrate` + `--drop-legacy` close the grace period cleanly. Downstream consumers (urgent lane, msg_claim broadcasts) updated.

**Ship it.** Sets up PR-5 (presence) and PR-6 (channels) with a clean foundation.

PR #61, opened 2026-04-21
feat(msg): presence / heartbeat — design PR-5

piekstra:
## Summary

PR-5 of `docs/designs/agent-messaging.md` (§2.6): per-handle presence files
(`<mailbox>/presence/<handle>.json`) with a `twapp msg presence` subcommand
group (`heartbeat`, `list`, `get`, `clear`). Coordinator-facing `list --stale`
filters by the 5× poll_interval_sec dormancy threshold; `clear` is the
offboard cleanup.

Heartbeating is the worker's responsibility today — skill docs
(`agent-coordinator` §3, `spawn-agent` worker protocol) now require it on
each `/loop` cycle and a `presence clear` on offboard. Automating heartbeats
inside the `/loop` skill itself is deferred per the briefing.

## What changed

- New module `src-tauri/src/cli/msg_presence.rs`:
  - `PresenceFile` data model (handle, status, last_heartbeat RFC3339,
    current_task, inbox_cursor, poll_interval_sec, claims).
  - `PresenceStatus` enum (`processing` | `idle` | `dormant`).
  - Path helpers (`presence_dir`, `presence_file`).
  - IO (`read_presence`, `write_presence`, `list_presence`).
  - `is_dormant()` — `last_heartbeat < now - 5 × poll_interval_sec`;
    corrupt timestamps treated as dormant so they don't hide from `--stale`.
  - CLI entrypoints (`cmd_heartbeat`, `cmd_list`, `cmd_get`, `cmd_clear`).
  - `merge_heartbeat()` preserves prior fields when a flag is omitted, so
    agents can heartbeat with just `--task` without rewriting claims or
    cursor each cycle.
- Wired a `Presence { command: PresenceCommands }` variant into
  `MsgCommands` in `src-tauri/src/cli/msg.rs` and dispatch in
  `src-tauri/src/cli/mod.rs`.
- `README.md` — new "Presence + heartbeats (PR-5)" subsection under the
  co-lab messaging section; PR-5 added to the coverage list.
- `skills/agent-coordinator/SKILL.md` — new §3 "Presence / heartbeats"
  subsection, heartbeat/clear lines added to the briefing-template
  Protocol block and to the worker offboard ritual.
- `skills/spawn-agent/SKILL.md` — worker Protocol example and Example A
  briefing updated to include the heartbeat + clear on offboard.

## Dormancy semantics

- **processing / idle** — file exists, `last_heartbeat` within
  `5 × poll_interval_sec`. Alive.
- **dormant** — file exists but past the threshold. Surfaced by
  `presence list --stale`. A worker deep in a write cycle can legitimately
  go dormant; senders use it as a hint, not a verdict.
- **dead** — no presence file at all ("never started or fully offboarded").
  `presence list` omits dead handles entirely.

## Test plan

Unit tests in `msg_presence.rs::tests` cover every ask in the briefing:

- [x] `heartbeat_creates_file_on_first_call`
- [x] `heartbeat_overwrites_existing` (advances timestamp, preserves
      unset fields)
- [x] `list_returns_all_files`
- [x] `list_stale_filters_by_threshold` (threshold scales with the file's
      own `poll_interval_sec`)
- [x] `get_unknown_handle_errors` (non-zero exit)
- [x] `clear_removes_file` (idempotent)
- [x] `dormant_threshold_uses_poll_interval` (live at 299s, stale at
      301s with interval=60; 1000s still live with interval=300; corrupt
      timestamp → dormant)
- [x] `merge_heartbeat_preserves_fields_when_flag_omitted`
- [x] `parse_claims_splits_and_trims`
- [x] `list_ignores_unparseable_json`

Also verified manually against a temp mailbox:

- [x] `heartbeat` without a prior file — defaults status=processing,
      interval=90.
- [x] Second `heartbeat --status idle --task "<s>"` preserves interval
      and claims, advances timestamp.
- [x] `list` (pretty + `--format json`) and `get <handle>`.
- [x] `get` on an unknown handle returns exit 1.
- [x] `clear` removes the file; second `clear` on the same handle still
      exits 0.

- Channels (PR-6).
- Automated heartbeats inside the `/loop` skill (later skill update).
- UI fleet pane (wave-3 consumer).

### DIFF
--- a/README.md
+++ b/README.md
@@ -575,12 +575,13 @@ back. See
 [`docs/designs/agent-messaging.md`](docs/designs/agent-messaging.md) for
 the full model. Today's CLI covers **PR-1** (scaffolding), **PR-2**
 (threading), **PR-3** (directory split + read cursors), **PR-4**
-(priority lane), and **PR-7** (archive rotation) of that design. Sends
-land under a per-recipient layout (`inbox/broadcast/`,
-`inbox/direct/<handle>/`, `inbox/channel/<name>/`); every send also
-drops a grace-period symlink at the flat `inbox/<filename>.md` path so
-un-upgraded readers doing `ls inbox/` keep working. Presence and
-first-class channel subscriptions still land in later PRs.
+(priority lane), **PR-5** (presence / heartbeat), and **PR-7** (archive
+rotation) of that design. Sends land under a per-recipient layout
+(`inbox/broadcast/`, `inbox/direct/<handle>/`, `inbox/channel/<name>/`);
+every send also drops a grace-period symlink at the flat
+`inbox/<filename>.md` path so un-upgraded readers doing `ls inbox/`
+keep working. First-class channel subscriptions still land in a later
+PR.
 
 #### Configure the mailbox
 
@@ -736,6 +737,73 @@ twapp msg ack 01JS4M7Q8W --note "rebasing now"
 ```
 
 `--mark-read` is opt-in; a plain `fetch` never advances the cursor.
+
+#### Presence + heartbeats (PR-5)
+
+`twapp msg presence heartbeat` writes `<mailbox>/presence/<handle>.json`,
+overwritten in place each call. The coordinator (and any curious peer)
+can see who's alive, what they're doing, and how long it's been since
+their last tick without opening their inbox. Design §2.6.
+
+```json
+{
+  "handle": "implementer-a",
+  "status": "processing",
+  "last_heartbeat": "2026-04-21T12:00:00Z",
+  "current_task": "rebasing onto main",
+  "inbox_cursor": "20260420T202845Z-9f2c1a",
+  "poll_interval_sec": 90,
+  "claims": ["channel:reviewers-standby"]
+}
+```
+
+```bash
+# Basic heartbeat — defaults handle to the current session name, status
+# to processing, poll_interval_sec to 90 on the first call (preserved on
+# subsequent calls if --interval is omitted).
+twapp msg presence heartbeat
+
+# Update just the status line; other fields are preserved.
+twapp msg presence heartbeat --task "rebasing onto main"
+twapp msg presence heartbeat --status idle --task "waiting on review"
+
+# Multi-field refresh.
+twapp msg presence heartbeat \
+  --status processing \
+  --task "writing tests" \
+  --interval 90 \
+  --claims channel:reviewers-standby,channel:announcements
+
+# Fleet view — every handle with a presence file.
+twapp msg presence list
+twapp msg presence list --format json | jq
+
+# Who's dormant? Shows only handles whose last_heartbeat is older than
+# 5 × poll_interval_sec.
+twapp msg presence list --stale
+
+# Single-handle lookup.
+twapp msg presence get implementer-a
+
+# On offboard, clear your file so peers don't see you as dormant.
+twapp msg presence clear
+```
+
+Three liveness states:
+
+- **processing / idle** — presence file exists and `last_heartbeat` is
+  within `5 × poll_interval_sec`. Alive.
+- **dormant** — presence file exists but `last_heartbeat` is past the
+  threshold. Surfaced by `presence list --stale`. A busy worker deep in
+  a write cycle can legitimately go dormant; senders use it as a hint,
+  not a verdict.
+- **dead** — no presence file at all. Means "never started or fully
+  offboarded". `presence list` omits dead handles entirely.
+
+Agents are expected to heartbeat on their /loop cadence (today, manually;
+a follow-up PR will wire heartbeats into the /loop skill itself). The
+`spawn-agent` and `agent-coordinator` skill docs call this out in the
+worker protocol.
 
 #### Migrating the inbox layout
 
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -109,6 +109,7 @@ operation — worker contract" below), self-merge criteria,
 offboard shape, hard rules.>
 
 - Invoke `/loop` after hello; poll every 90-120s. Never finish a turn asking the user a question — mailbox the coordinator if stuck.
+- Heartbeat each `/loop` cycle: `twapp msg presence heartbeat --task "<one-line current step>"`. On offboard, `twapp msg presence clear`. See §3 Presence / heartbeats.
 
 ## Coordinate with
 <Names of other in-flight workers whose work overlaps —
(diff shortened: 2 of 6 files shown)


piekstra (commented):
**Ship it.** Right-scoped §2.6 implementation; `is_dormant` math is careful; field-preservation via `merge_heartbeat` is the right ergonomic; skill docs instrument both the template protocol block and the §3 operator manual so future briefings inherit the heartbeat mandate.

## Policy scan

Clean.

## What's right

### Data model + IO

- **`PresenceFile`** matches design §2.6 field-for-field (`handle`, `status`, `last_heartbeat`, `current_task`, `inbox_cursor`, `poll_interval_sec`, `claims`). Optional fields use `#[serde(default, skip_serializing_if = "...")]` so legacy JSON without `current_task` / `inbox_cursor` still parses, and no JSON bloat when fields are absent. ✓
- **`PresenceStatus` enum** (`processing`/`idle`/`dormant`) is `ValueEnum` for clap + `lowercase`-renamed for JSON. The `Default` derive picks `Processing` which matches design's implicit "new agent is working" baseline.
- **`read_presence` returns `Ok(None)` on NotFound** — absence is the expected case, not an error. Parse errors propagate (with path context) so corrupt files are loud, not silent. ✓
- **`write_presence` creates the directory on demand** (`create_dir_all`). First heartbeat for a handle needn't know the parent already exists.
- **`list_presence` skips unparseable files** with a `log::debug!`. Stable sort by handle name. Matches the "filesystem state is authoritative, CLI is convenience" posture from the messaging design.

### Dormancy math

`is_dormant` is the load-bearing function; it's careful in three places:

1. **Corrupt `last_heartbeat` → dormant** (not "fresh" / not "skip"). Right call — a malformed timestamp shouldn't hide a handle from `--stale`; that's how stuck workers stay invisible.
2. **Negative age (writer's clock ahead of ours) → not dormant**. Tolerates modest clock skew between hosts on a shared filesystem. ✓
3. **`saturating_mul` on `DORMANT_MULTIPLIER * poll_interval_sec`** + `.max(1)` on the interval → no overflow for pathological intervals, no divide-by-zero on `0`. Defensive without being paranoid.

The dormancy threshold (5 × poll_interval_sec) is pulled into a `const DORMANT_MULTIPLIER = 5` constant with a design-doc reference, so a future change stays centralized.

### `merge_heartbeat` — the right ergonomic

This is the single most important UX choice in the PR. By merging CLI flags over the prior file (rather than replacing the whole record), agents can:

- `twapp msg presence heartbeat --task "writing tests"` — updates only the task.
- `twapp msg presence heartbeat` — re-heartbeats with current fields, just refreshes `last_heartbeat`.
- `twapp msg presence heartbeat --status idle --task "waiting on review"` — multi-field update, but untouched fields (interval, claims, cursor) preserved.

A worker doesn't need to re-send `--interval 90 --claims channel:foo,channel:bar` on every heartbeat. Test `merge_heartbeat_preserves_fields_when_flag_omitted` pins this behavior.

Separate from `cmd_heartbeat` for unit-testability without filesystem setup. Right separation.

### Status-vs-liveness distinction

Three states, each handled separately:

- **`status` field** (processing/idle/dormant) is self-reported by the worker. A worker can explicitly set `--status dormant` to say "I'm stepping away briefly."
- **Derived dormancy** (`is_dormant`) is computed from `last_heartbeat` age, not from the `status` field. A worker that crashes without updating status just shows as stale.
- **Dead** (no file) is omitted from `list` entirely. The PR body calls this out: "dead ≠ dormant; absent file means never started OR fully offboarded".

This is exactly the design §2.6 framing. The skill doc makes this distinction visible to coordinators.

### CLI surface

- `heartbeat` / `list` / `get` / `clear` — minimum viable verb set. No flags that imply future features (e.g. no `watch` mode, no `--format yaml`, no subscription API).
- `--stale` filter on `list` is the coordinator's main query. `list_stale_filters_by_threshold` test pins the filter behavior with a fresh and a stale file.
- Empty-output messages diverge by filter: `(no stale presence files)` vs `(no presence files)` — tiny UX win, scripters still prefer `--format json`.
- `clear` is the offboard primitive. Absent `--handle` defaults to the current session name. Test `clear_removes_file` verifies it.

### Skill/doc updates

- **agent-coordinator §2 template Protocol block** gains one line: `twapp msg presence heartbeat --task "<one-line>"` + `twapp msg presence clear` on offboard. The briefing-template inline pattern matches #44/#47's approach — future briefings copy-paste the template and inherit the heartbeat contract automatically.
- **agent-coordinator §3 "Presence / heartbeats"** new subsection explains the three states, the `--stale` workflow, and the pair-with-reap-sweep note. Exactly right placement — next to §3 mailbox protocol, so a coordinator reading the skill finds everything about the bus in one section.
- **spawn-agent worker protocol** gains the heartbeat+clear lines. Forward-compat: when the /loop skill eventually automates heartbeats (per the PR body), the briefing template line becomes redundant; easy to strip.
- **README** section documents the feature end-to-end with a sample JSON and every CLI verb.

## Non-blocking nits

- **`parse_claims`** splits on `,`, trims whitespace, filters empty. No quoting story — a claim string containing a literal `,` (e.g. a channel name with embedded punctuation) would be split. Unlikely in practice (channel names follow the handle-format convention); worth a note in the `--claims` help text if that convention ever loosens.
- **`heartbeat` doesn't validate `inbox_cursor`** shape. An agent could pass garbage and it'd just survive on disk. The field is informational (the cursor fast-path per design §2.7's "fast path" mention), so garbage here doesn't corrupt anything. Tolerable.
- **`is_dormant` doesn't distinguish "very old file" from "just past threshold".** `list --stale` surfaces both as a single "dormant" bucket. A future `--stale-age-seconds <N>` could add a staler-than-stale filter for ops visibility; non-blocking for now.
- **`current_task` is free-text.** No size cap. If an agent writes a 10KB task description it'll bloat the JSON but not break anything. Worth a one-line comment or a 256-char clip in `merge_heartbeat` if this ever shows up as noise.
- **The `presence/<handle>.json` file is overwritten-in-place** (per design §2.6), not tmp+rename. A reader who happens to catch the write mid-flight will get a parse error and get logged at debug. Consistent with the other overwrite-in-place files in the design; worth leaving alone, but a future tmp+rename upgrade would eliminate the 10-ms race window.

## Test coverage

10 tests: first-call, overwrite-existing, list returns all, list --stale filter, get unknown errors, clear removes, dormancy-threshold math, merge-field-preservation, claims parser, unparseable-json skipped. Covers every public path. ✓

## Scope discipline

- PR body explicitly names what's deferred: automating heartbeats into the `/loop` skill itself.
- No touches to the `/loop` skill runtime or to Tauri GUI surfaces — the CLI is the only interface this PR adds. GUI integration lives in a separate follow-up (likely wave 3 UI).
- No changes to the presence schema beyond design §2.6; `claims` field is present but only used for display. PR-6 (channels, first-class subscriptions) will consume it.

## Status

- CI: SUCCESS.
- `mergeStateStatus: CLEAN`.

## Summary

Focused design §2.6 implementation; careful dormancy math; field-preserving merge behavior removes friction for every-/loop heartbeat calls; skill and template updates propagate the heartbeat mandate to future briefings. Ship it. Sets up PR-6 (channels) with a `claims` field already in the record.

PR #62, opened 2026-04-21
feat(ui): coordinator fleet-pane quick-action context menu

piekstra:
## Summary

Implements **§3.5** of `docs/designs/agent-aware-ui.md` — the per-agent context menu that the fleet pane hangs on each agent row. The fleet-pane PR (parallel wave-3 work) will import `AgentContextMenu` and wire it to its row components; this PR ships the component, the Tauri commands behind each action, tests, and docs. Composer flow is unchanged — new priority presets open the existing composer via an `onOpenComposer` callback.

### What ships

- `AgentContextMenu` component (portal-based, keyboard nav, viewport clamping, per-agent gating per §3.6)
- **Open window** — `focus_agent_window` Tauri cmd: `open -a` the running instance `.app`, or report the session offline so the item dims
- **Send direct / urgent / blocker** — opens the existing composer with `to: <handle>` + `priority` preselected
- **View recent activity** — `fetch_agent_activity` Tauri cmd: shells `twapp msg fetch --all --format json`, filters in-process for `from == handle || handle ∈ to || handle ∈ cc` (CLI `--for` is inbound-only)
- **View PR activity** — `list_agent_prs` Tauri cmd: `gh pr list --author <handle> --json ...`; missing `gh` surfaces as a typed skip message in the modal
- **Stop agent** — `stop_agent` Tauri cmd (`twapp stop --name <handle>` + optional `--force`). Fronted by a confirmation dialog; warns on unshipped PR; hidden for coordinator / self rows
- Input validation rejects handles with shell metacharacters / whitespace / path separators (defense in depth)

### Tests

- 13 vitest cases — menu gating (Stop hidden for coordinator / self, Open disabled on self), destructive styling, blocker hint text, anchor clamping math
- 12 cargo tests — handle validation, `build_stop_argv`, `build_gh_pr_argv` (JSON field coverage), `parse_gh_pr_stdout` (happy + non-array + empty), activity filter (inbound, outbound, cc, self-broadcast, skip others-to-all), `sort_and_truncate`
- Full suites green: **140/140** vitest, **244/244** cargo

### Out of scope

- No new backend tooling — menu items shell existing commands (per briefing)
- Multi-select / bulk actions — future
- Consumer wiring — fleet-pane PR imports `AgentContextMenu` when it lands

## Test plan

- [x] `npm test` — 140 pass
- [x] `cargo test --lib` — 244 pass
- [x] `npm run build` — clean production bundle
- [x] `cargo clippy --lib --no-deps` — no new warnings from `agent_actions.rs`
- [ ] Manual smoke with live fleet pane (blocked on fleet-pane PR landing)

Reviewer: `twapp-reviewer` / `twapp-reviewer-2`. Ship-on-CLEAN + Ship it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -727,6 +727,18 @@ collapses after the queue has been empty for a minute. Click a row to
 open a read-only message view. Blockers get the strongest red accent,
 urgents a muted one. Single-session users with no handle see no panel.
 
+#### Per-agent quick actions (context menu)
+
+Right-clicking a fleet-pane row (and later any co-lab row) opens a
+context menu with the most common coordinator actions: raise the
+agent's window, send a direct / urgent / blocker message (opens the
+composer prefilled), view recent mailbox activity, view their PR
+activity via `gh pr list --author <handle>`, and stop the agent via
+`twapp stop --name <handle>`. The stop flow ships with an explicit
+confirmation dialog and an optional SIGKILL escalation checkbox; it's
+hidden when the row targets the coordinator or the user's own session.
+See `docs/designs/agent-aware-ui.md` §3.5 for the full spec.
+
 #### Reading legacy (bare) files
 
 `fetch` accepts both the new fenced-frontmatter shape and the older bare
--- a/docs/designs/agent-aware-ui.md
+++ b/docs/designs/agent-aware-ui.md
@@ -404,6 +404,11 @@ The button is hidden (never rendered) when:
 
 ### 3.5 Quick actions from the dashboard
 
+> **Landed** (ui-quick-actions PR): the `AgentContextMenu` component +
+> supporting Tauri commands (`focus_agent_window`, `stop_agent`,
+> `list_agent_prs`, `fetch_agent_activity`) ship this shape. The fleet-pane
+> PR wires the menu to the per-agent rows it introduces.
+
 Right-click on any fleet-pane row opens a context menu. Keyboard: `⌘K`
 with a row focused. Items:
 
--- a/src-tauri/src/gui/agent_actions.rs
+++ b/src-tauri/src/gui/agent_actions.rs
@@ -1,0 +1,438 @@
+//! Tauri commands backing the fleet-pane per-agent context menu
+//! (`src/components/AgentContextMenu.tsx`). Each command is a thin shell
+//! around an existing CLI:
+//!
+//! - `focus_agent_window` — `open -a ~/.config/twapp/instances/<handle>.app`
+//!   when the instance is running; otherwise reports the session as offline
+//!   so the UI can dim the menu item.
+//! - `stop_agent` — `twapp stop --name <handle>` (with `--force` escalation).
+//! - `list_agent_prs` — `gh pr list --author <handle> --json …`. Missing `gh`
+//!   surfaces as a typed `AgentActionError::GhMissing` so the UI can skip the
+//!   item per the §3.5 spec.
+//! - `fetch_agent_activity` — `twapp msg fetch --all --format json`, then
+//!   filtered in-process for messages to/from/cc'ing the handle so "recent
+//!   activity" spans both directions. The CLI's `--for` filter only covers
+//!   inbound messages, so we do the from-side filter here rather than
+//!   shelling twice.
+
+use super::msg::{parse_fetch_stdout, FetchedMessage};
+use super::sessions::{check_instance_running, sanitize_instance_name};
+use super::types::*;
+use serde::{Deserialize, Serialize};
+use std::process::{Command, Stdio};
+
+/// Arguments for the stop-agent flow. `force` escalates to SIGKILL on the
+/// CLI side if the graceful SIGTERM doesn't land.
+#[derive(Debug, Clone, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct StopAgentArgs {
+    pub handle: String,
+    #[serde(default)]
+    pub force: bool,
+}
+
+/// Outcome of a focus attempt. `false` means the instance isn't running and
+/// the menu item should be treated as a no-op rather than an error.
+#[derive(Debug, Clone, Serialize)]
+pub struct FocusResult {
+    pub focused: bool,
+    /// Absolute path to the instance bundle, or empty when not running.
+    pub app_path: String,
+}
+
+/// One row in the "View PR activity" modal. Shape mirrors the `gh pr list
+/// --json` fields we request, so adding more columns is a CLI-flag change.
+#[derive(Debug, Clone, Serialize)]
+pub struct AgentPr {
+    pub number: u64,
+    pub title: String,
+    pub state: String,
+    pub url: String,
+    pub updated_at: String,
+    pub is_draft: bool,
+}
+
+/// Validate a handle before shelling out. Rejects empty and anything that
+/// isn't a plausible session name so we never splat shell metacharacters
+/// into argv (the CLIs themselves don't shell-interpret, but defense in
+/// depth: `.`, `/`, whitespace, quotes all fail here).
+pub fn validate_handle(handle: &str) -> Result<&str, String> {
+    let trimmed = handle.trim();
+    if trimmed.is_empty() {
+        return Err("Agent handle is required".into());
+    }
+    let ok = trimmed
+        .chars()
+        .all(|c| c.is_alphanumeric() || c == '-' || c == '_');
+    if !ok {
+        return Err(format!(
+            "Invalid agent handle (letters / digits / - / _ only): {}",
+            trimmed
+        ));
+    }
+    Ok(trimmed)
+}
+
+fn twapp_binary() -> std::path::PathBuf {
+    std::env::current_exe().unwrap_or_else(|_| std::path::PathBuf::from("twapp"))
+}
+
+fn instance_app_path(handle: &str) -> Option<std::path::PathBuf> {
+    let instances = dirs::home_dir()?.join(".config/twapp/instances");
+    let safe = sanitize_instance_name(handle);
+    Some(instances.join(format!("{}.app", safe)))
+}
+
+#[tauri::command]
+pub fn focus_agent_window(handle: String) -> Result<FocusResult, String> {
+    let handle = validate_handle(&handle)?;
+    if !check_instance_running(handle) {
+        return Ok(FocusResult { focused: false, app_path: String::new() });
+    }
+    let app = instance_app_path(handle).ok_or_else(|| "No home directory".to_string())?;
+    Command::new("open")
+        .args(["-a", &app.to_string_lossy()])
+        .spawn()
+        .map_err(|e| format!("Failed to focus {}: {}", app.display(), e))?;
+    Ok(FocusResult {
+        focused: true,
+        app_path: app.to_string_lossy().to_string(),
+    })
+}
+
+/// Build the argv for `twapp stop --name <handle> [--force]`. Pure so the
+/// flag wiring (and omission of `--force` when false) is unit-testable.
+pub fn build_stop_argv(handle: &str, force: bool) -> Vec<String> {
+    let mut argv = vec![
+        "stop".to_string(),
+        "--name".to_string(),
+        handle.to_string(),
+    ];
+    if force {
+        argv.push("--force".to_string());
+    }
+    argv
+}
+
+#[tauri::command]
+pub fn stop_agent(
+    args: StopAgentArgs,
+    config: tauri::State<'_, GuiArgs>,
+) -> Result<(), String> {
+    let handle = validate_handle(&args.handle)?;
+    let argv = build_stop_argv(handle, args.force);
+    let bin = twapp_binary();
+    let mut cmd = Command::new(&bin);
+    cmd.args(&argv).stdout(Stdio::piped()).stderr(Stdio::piped());
+    if let Some(d) = config.cwd.as_deref() {
+        cmd.current_dir(d);
+    }
+    let out = cmd
+        .output()
+        .map_err(|e| format!("Failed to launch {}: {}", bin.display(), e))?;
+    if !out.status.success() {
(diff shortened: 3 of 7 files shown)


piekstra (commented):
**Ship it.** Per-agent context menu ships as a standalone component + clean Tauri commands; the in-isolation approach (pane wires the menu in a follow-up) is the right scoping; `validate_handle` is defense-in-depth on every shell-out; 12 tests pin the argv/parse/filter contracts.

## Policy scan

Clean.

## What's right

### Scope — standalone-component-plus-commands

PR body frames it correctly: "The fleet-pane PR (parallel wave-3 work) will import `AgentContextMenu` and wire it to its row components; this PR ships the component, the Tauri commands behind each action, tests, and docs." Ship the component + backend commands now, wire them into #63's pane in a follow-up. That's better than racing against #63's base branch — the context-menu component can be reviewed and merged in isolation without blocking on the fleet-pane shape.

### `validate_handle` — the right pre-shell check

Every command calls `validate_handle` first. Only `alphanumeric | - | _`. Rejects empty, whitespace-only, and any handle with `/`, `.`, quotes, shell metacharacters. Defense in depth — the downstream CLIs don't shell-interpret argv, but this closes the whole surface in one place.

- `stop_agent` — validates before building the `twapp stop --name <handle>` argv.
- `list_agent_prs` — validates before `gh pr list --author <handle>`.
- `focus_agent_window` — validates before `open -a <app>`.
- `fetch_agent_activity` — validates before the in-process filter.

`validate_handle_rejects_metacharacters` test names the hostile inputs (`../escape`, `a b`, `a;b`, `a'"`, etc.).

### Pure builders / parsers

Each shell-out is paired with a pure helper:

- `build_stop_argv(handle, force)` — unit-testable argv shape. Verifies `--force` threading.
- `build_gh_pr_argv(handle, limit)` — pins the `--json` field set so downstream `parse_gh_pr_stdout` stays aligned.
- `parse_gh_pr_stdout` — handles empty / non-array / ignores unknown fields. 3 tests (two PRs, empty, non-array rejection).
- `filter_activity_for(handle, messages)` — in-process filter for "agent touched this, in either direction". Deliberately excludes inbound broadcasts from other handles (correct — they'd flood the activity view). Test covers inbound, outbound, cc, and self-broadcast cases.
- `sort_and_truncate(messages, limit)` — newest-first; `limit > len` stays as-is. Test `sort_and_truncate_limit_larger_than_input` pins this.

### `fetch_agent_activity` design

Pulls `--limit = limit * 4, min 80` messages from the CLI and filters post-hoc to the handle. Rationale in the docstring:

> "a recent burst of broadcasts from other handles can otherwise crowd out the agent's own messages"

That's right. The CLI's `--for` filter only sees inbound (as the docstring says), so for "recent activity about handle X" we need both directions, done by filtering JSON in-process. Pulling a 4× buffer is a reasonable middle between "one shell call" and "always see 20 of X's messages."

### `list_agent_prs` with `gh missing`

`Command::new("gh").spawn()` with `NotFound` mapped to the stable string `"gh CLI not installed"`. Frontend can skip the menu item without parsing a shell error. Matches design §3.5's "skipped if `gh` not configured".

### `focus_agent_window` graceful offline

Returns `FocusResult { focused: false, app_path: "" }` when the instance isn't running. Not an error — the UI dims the menu item. `check_instance_running` is the existing launcher helper (reused, not reinvented). ✓

### Design doc update

§3.5 gets a **"Landed"** callout (not "partially" — this PR ships the full §3.5 surface per the spec, minus the pane-wiring which lives in #63's follow-up). Honest about scope. ✓

## Non-blocking nits

- **`fetch_agent_activity` 4× buffer heuristic** is fine for small mailboxes, but if a fleet generates 1000+ messages/day the 4× might still not surface older activity from a quiet handle. Consider accepting `--since <ts>` to bound the query instead of relying purely on `--limit`. Non-blocking for v1.
- **`validate_handle` rejects `.`** — that's stricter than the CLI's own handle validation (which allows dotted handles via `handle.sub`). If any existing handle uses dots, this command rejects it even though the rest of twapp accepts it. Conservative; worth a future relaxation if the use case shows up.
- **`AgentPr::is_draft`** defaults to `false` on missing field. The `gh pr list --json isDraft` output is always present (per gh's schema), so the fallback is defensive, not reachable. Fine.
- **`stop_agent` return type is `Result<(), String>`** — no stdout surfaced to the caller. Success shows nothing; failure shows stderr. That matches the confirmation-dialog UX (the user already confirmed, now just run).

### Test coverage (12 tests)

Validation (3) + argv (2) + gh-pr parsing (3) + activity filter + sort-truncate (2). No end-to-end test of `fetch_agent_activity` because it requires a real `twapp msg fetch` shell-out + mailbox; the pure helpers are dense enough.

## Summary

Right scope (ships the component + commands; pane wires them up later), right defensive posture (validate → build-argv → shell-out → parse), right test coverage on every pure function. Design doc §3.5 cleanly marked landed.

**Ship it.** When #63 merges (or after this merges first — either order works since neither imports the other's types), a small follow-up PR to `FleetPane.tsx` imports `AgentContextMenu` and wires `onContextMenu` on each row.

piekstra (commented):

## Policy scan

## Correctness

### Backend — `gui/agent_actions.rs`

**`validate_handle`** — input validation is the thing worth getting right in a shell-out layer:
- Trims, rejects empty, restricts to `[A-Za-z0-9_-]`. That rejects shell metacharacters, whitespace, `/`, `.` path traversal, quotes. `validate_handle_rejects_metacharacters` exercises four of the foot-gun cases (`foo; rm -rf /`, space, `../etc/passwd`, `quote"`). Good defense-in-depth even though the callees don't shell-interpret.
- Called at the top of every Tauri command that takes a handle. Zero paths bypass it.

**`focus_agent_window`** — uses `check_instance_running` + `open -a <instance>.app`. When not running, returns `FocusResult { focused: false, app_path: "" }` rather than erroring — the UI can surface "offline" or dim the menu item without a modal. Right disposition: "session is down" is a state, not an error.

**`stop_agent` + `build_stop_argv`** — separate pure builder for testability. `stop_argv_with_force` / `stop_argv_minimal` lock both branches. No `force: bool` coupling in argv (the flag is appended conditionally). Good.

**`list_agent_prs` + `build_gh_pr_argv` + `parse_gh_pr_stdout`** —
- `gh_argv_requests_expected_json_fields` asserts the six field names required by `AgentPr` are in `--json`. That's the contract pin that'll save someone from silently breaking the modal by removing a field.
- `parse_gh_pr_stdout` tolerates missing optional fields (`state`, `url`, `updatedAt` default to `""`; `isDraft` defaults to `false`). Rejects non-array. `parse_gh_pr_stdout_rejects_non_array` tests that.
- **Missing-`gh` detection**: `io::ErrorKind::NotFound` surfaces as the stable prefix `"gh CLI not installed"` so the UI can skip the item per §3.5's spec. That's a clean typed-error substitute — worth a short sentence in the doc so UI authors know to grep for that prefix.
- Hard-coded `--limit 5` default (overridable via `Option<u32>`) is reasonable for a modal; users who want more can hit the GitHub URL.

**`fetch_agent_activity`** — this is the subtle one because `twapp msg fetch --for <h>` is INBOUND-only. To get outbound messages (the agent sending to others), the Rust layer pulls `--all` and filters in-process:
- `filter_activity_for` includes `from == handle`, `to contains handle`, `cc contains handle`. The explicit "broadcasts from OTHERS don't count" exclusion is the right call — otherwise the activity feed for `alice` would be 90% noise from every broadcast in the mailbox. `filter_activity_includes_inbound_outbound_cc_and_self_broadcasts` pins all three inclusions and both exclusions (others-to-all, unrelated-pairs). Good coverage.
- `sort_and_truncate` sorts by ts descending, truncates to `limit`. `sort_descending_and_truncate` + `sort_and_truncate_limit_larger_than_input` cover both branches.
- `fetch_limit = (limit.saturating_mul(4)).max(80)` — pulls a 4× window so a burst of other-handle broadcasts can't crowd out the agent's own messages. That's the right instinct. If someone sets `limit: 100` this'd fetch 400 — fine, but worth a cap (say `min(fetch_limit, 2000)`) so a malformed caller can't accidentally request all of history.

### Frontend — `AgentContextMenu.tsx`

**`buildMenuItems`** (pure, exported, unit-tested):
- 7 items in a stable order. Stop is the only one with hard visibility gating (hidden for coordinator + self per §3.5 + §3.6). Everything else stays visible, sometimes dimmed — matches the "menu shape is stable, muscle memory survives edge cases" principle. `preserves menu order when Stop is hidden so muscle memory survives` locks this explicitly.
- `disables Open window when target is self` — self-targeted Open is semantically meaningless; dimmed not hidden.
- `attaches a hint to Send blocker describing the recipient impact` — the `hint: "Asks the recipient to stop current work"` exists specifically to remind the coordinator what they're about to do. That's thoughtful.
- `Stop` has `destructive: true` → red tint (`.agent-menu-confirm-destructive` in CSS). Good.

**`clampAnchor`** — pure viewport clamping with a 4px pad. Tests cover mid-viewport (no-op), past-right, past-bottom, negative, AND "menu larger than viewport" (doesn't NaN). That last case is the kind of unit test that rarely exists but catches the bug people only find at a 1024x768 resolution. Nice.

**Composer integration** — blocker / urgent / direct all open the existing `MessageComposer` (from PR #50) with `to` + `priority` preselected via an `onOpenComposer` callback. No duplication of composer logic. Good reuse.

**Portal + keyboard nav** — the component uses `createPortal` (so the menu isn't clipped by a parent overflow:hidden) and mentions "keyboard nav" in the PR body. I didn't see keyboard handler tests in the diff (focus-up/down, Enter-to-activate, Esc-to-close) — those are the kinds of flows that silently regress. Not a blocker but worth a follow-up vitest or two with a fake-key dispatcher.

**Stop confirmation dialog** — confirmation modal with:
- "agent's unshipped PR" warning when `hasUnshipedPr` is set (orange `.agent-menu-warning` chrome)
- optional `--force` (SIGKILL) checkbox
- destructive red confirm button
- separate styling from the composer modal chrome

Right UX — stop is consequential enough to gate behind an explicit confirmation.

## Scope discipline
Files match the briefing: new `gui/agent_actions.rs` + `AgentContextMenu.{tsx,test.ts}` + handler registration in `gui/mod.rs` + CSS + README / design doc.

**Merge ordering with PR #63**: both PRs add a module line + handler entries to `src-tauri/src/gui/mod.rs`. The alphabetized order is:
- #62 adds `pub mod agent_actions;` (sorts to TOP of the module list)
- #63 adds `pub mod fleet;` (sorts between `files;` and `monitor;`)

Both are CLEAN today. Whichever lands second needs a 2-line rebase adding the other's module line + handler entry. Noted on #63 too.

**Consumer wiring**: the PR body explicitly says "fleet-pane PR imports `AgentContextMenu` when it lands" — so today's #62 ships the component + tests without a consumer. That's a reasonable split; the wiring is a 5-line change in #63's successor and keeps this PR focused.

## Tests
- Rust: 12 new (handle validation x3, stop_argv x2, gh argv + parse x3, activity filter + sort x2+2). 244/244 pass per PR body. The activity-filter test is especially good — it asserts five separate inclusion/exclusion cases in one fixture.
- TS: 13 new (menu gating x7, anchor clamping x5, plus order-preservation). 140/140 pass per PR body.

Gaps (not blockers):
- No test for the stop-agent confirm modal's `hasUnshipedPr` warning rendering — a vitest that builds the modal props with `{ hasUnshipedPr: true }` and asserts the `.agent-menu-warning` class appears would lock the "don't silently drop the warning" invariant.
- No test for the "`gh` not installed" message-propagation path — assert that `Err("gh CLI not installed")` from the Rust side bubbles through `list_agent_prs` and hits the UI's skip-branch instead of an error modal.
- No test for keyboard navigation (focus-wrap on arrow keys, Enter activates highlighted item, Esc closes). The PR body mentions keyboard nav exists; it's just not test-pinned.

## Docs consistency
- README "Per-agent quick actions (context menu)" subsection gives the right operator-facing description: lists the seven items, explains the `gh pr list` path, and flags the confirmation dialog + SIGKILL escalation. Mentions the hide-for-coordinator + self gating.
- `docs/designs/agent-aware-ui.md` §3.5 gets a "Landed (ui-quick-actions PR)" callout at the top. Matches the pattern #63 used for §3.3.

- `fetch_agent_activity`'s `fetch_limit = max(80, limit * 4)` has no upper cap. A caller that sends `limit: 10000` triggers a 40000-message fetch; `msg fetch` can handle that, but the modal truncates to the requested 10000 anyway. A `.min(2000)` upper bound would prevent the pathological case without affecting real-world usage.
- `list_agent_prs` hard-codes `--state all`. If the modal mostly cares about open+merged PRs, `--state merged,open` would avoid showing stale closed PRs. Matter of product taste.
- The `"gh CLI not installed"` prefix is a magic string. Consider exporting it as `pub const GH_NOT_INSTALLED: &str = "gh CLI not installed";` so UI and Rust share a constant, and `list_agent_prs` pin test can assert the exact prefix.
- `clampAnchor`'s 4px pad is a magic number — fine, but a `const PAD = 4;` with a brief comment on why 4 (not 0, not 12) would help a future theming pass.
- CSS: `.agent-menu-confirm-destructive:disabled` reads the same bg color as enabled (`#d14545`) with 0.6 opacity. If a user clicks rapidly and the button re-enables mid-animation, the flicker will be subtle. Low-prio.

Good PR. The handle-validation discipline, pure extraction of `buildMenuItems` + `clampAnchor` + `build_stop_argv` + `build_gh_pr_argv` + `parse_gh_pr_stdout` + `filter_activity_for` + `sort_and_truncate` into individually-testable helpers is exactly the right shape. The subtle bits (§3.6 stop-hidden gating, `--all` + post-hoc filter for activity, missing-`gh` stable-prefix surfacing, hasUnshippedPr warning in confirm) are all tested or intentionally documented. Merge-after-#63 is a 2-line alphabetical rebase; whichever lands first wins.

PR #59, opened 2026-04-21
feat(ui): co-lab window title prefix + coordinator tag in sessions list

piekstra:
## Summary

Ships the minimum **co-lab-identifying chrome** in the launcher and OS
window title, per briefing `twapp-colab-ui-chrome.md`. Everything
surfaces metadata already landed by #43 (`role` + `provenance`) and #54
(`colab_group`) — no new backend fields.

## Changes

- **Window title formatter** — new `src-tauri/src/gui/title.rs`
  centralizes `format_window_title(name, role, provenance)`:
  - `role` ∈ co-lab archetypes → `twapp - co-lab:<role> - <name>`
  - `provenance == "spawned"` → `twapp - co-lab - <name>`
  - everything else → unchanged `twapp - <name>` (or bare `twapp`)
  Wired in `gui/mod.rs` — the GUI reads `role` + `provenance` from the
  session file's `.twapp-session.json` at startup.
- **Sessions-list chip** — in `SessionLauncher.tsx`, co-lab-qualifying
  rows (role set OR spawned) now render a neutral `CO-LAB` chip in
  `launcher-session-meta`. Coordinator sessions trade the chip for a
  prominent `COORDINATOR` tag in the launcher accent color. Plain user
  sessions unchanged.
- **Helper** — `src/utils/colab.ts` exposes `isColabSession` and the
  archetype list, mirroring `gui::title::COLAB_ROLE_ARCHETYPES`.
- **Docs** — `agent-aware-ui.md` §3.1 / §3.2 flagged "partially
  landed" with pointers here; README's co-lab section grows a
  "Identifying a co-lab session at a glance" subsection.

## Tests

- `cargo test --lib` → 189/189 passing (10 new in `gui::title::tests`,
  one per branch including the full archetype loop and the two
  non-archetype edge cases).
- `npm test` → 127/127 passing (9 new in `src/utils/colab.test.ts`).
- `npx tsc --noEmit` → clean.

## Test plan

- [ ] `cargo test --lib` green
- [ ] `npm test` green
- [ ] Launch a plain user session (`twapp work --name bash`) → window
      title reads `twapp - bash`; launcher row has no CO-LAB chip.
- [ ] Launch a spawned session with no role (e.g. `--from-file`) →
      window title reads `twapp - co-lab - <name>`; launcher row shows
      a neutral `CO-LAB` chip.
- [ ] Launch a session with `role: "coordinator"` → window title reads
      `twapp - co-lab:coordinator - <name>`; launcher row shows the
[...]

PR #63, opened 2026-04-21
feat(ui): coordinator fleet pane (presence-aware)

piekstra:
## Summary

- New sidebar **Fleet** pane that activates when the session declares `role: "coordinator"`. Lists every handle with a live `<mailbox>/presence/<handle>.json`, with status dot, role/provenance chips, unread + urgent counts, heartbeat age, and truncated `current_task`. Polls `list_fleet` every 5s; row-click raises the target twapp window via the existing `launch_session` focus path.
- New `list_fleet` Tauri command in `src-tauri/src/gui/fleet.rs` that reads presence files, counts `inbox/direct/<handle>/` + `inbox/urgent/<handle>/` + `inbox/urgent/all/`, decorates each row with the launcher's session-on-disk metadata, and (optionally) scopes to a `colab_group`.
- Implements design §3.3 (read-only fleet pane item from the roadmap in §5 item 3). Dashboard-mode expansion, inbox/broadcast panes, spawn timeline, and quick-action context menu remain separate wave-3 PRs per the design doc.

## What lands

- `src-tauri/src/gui/fleet.rs` — `list_fleet` command + pure `build_fleet_rows` builder + `count_files_in` / `count_direct` / `count_urgent` helpers. 7 unit tests, no new deps.
- `src/components/FleetPane.tsx` — collapsible sidebar panel with status dots, role chips, provenance glyphs, urgent/unread chips, heartbeat age, and clickable rows. Hidden for non-coordinators.
- `src/components/FleetPane.test.ts` — 9 vitest tests for the pure helpers (age formatter, summarizer, dot routing, task truncation).
- `src/App.tsx` — loads `role` + `colab_group` from `get_session_info` at boot and renders `FleetPane` above `UrgentInbox`.
- `src/App.css` — fleet pane styles (dots, chips, row states, dormant opacity).
- Docs: `docs/designs/agent-aware-ui.md` §3.3 marked partially landed; `README.md` gains a "Coordinator fleet pane" subsection; `skills/agent-coordinator/SKILL.md` notes the UI pane alongside `twapp msg presence list`.

## Out of scope (per briefing + §3.3 notes)

- Dashboard-mode toggle that collapses the terminal.
- Inbox / Broadcast panes inside the dashboard.
- Spawn / offboard timeline.
- Quick-action context menu (open window / stop agent / send urgent / …).
- Cross-group view.

## Test plan

- [x] `cargo test --lib` — 239/239 pass (7 new fleet tests + unchanged suite).
- [x] `cargo clippy --lib` — no new warnings attributable to this change.
- [x] `npx vitest run` — 136/136 pass (9 new FleetPane tests + unchanged suite).
- [x] `npm run build` (tsc + vite) — clean.
- [ ] Manual smoke: coordinator with live workers → rows render sorted urgent-first; a dormant worker collapses to muted dot; row-click raises the worker's window. (Deferred: no active coordinator fleet to point this build at; the test suites cover the shape.)

## Dependency note

Messaging PR-5 (`feat(msg): presence / heartbeat`, #61) is the prerequisite and has merged. Automating heartbeats inside the `/loop` skill is a separate follow-up per the PR-5 offboard notes — until that lands, presence files are only present when an agent manually runs `twapp msg presence heartbeat`, so the pane legitimately shows "No workers online" in most repos today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -443,6 +443,37 @@ and the unchanged `twapp - <name>` for plain user sessions. The formatter
 lives in [`src-tauri/src/gui/title.rs`](src-tauri/src/gui/title.rs) with a
 unit test per branch.
 
+### Coordinator fleet pane
+
+A session whose `.twapp-session.json` declares `role: "coordinator"`
+renders a **Fleet** panel in its sidebar, above the notes / prompts
+stack. The pane polls the `list_fleet` Tauri command every 5 seconds and
+lists one row per handle with a live `<mailbox>/presence/<handle>.json`:
+
+- Status dot — green fill for `processing`, hollow for `idle`, muted for
+  `dormant` (derived from `last_heartbeat` aged past `5 × poll_interval_sec`
+  per the messaging design's dormancy rule).
+- Handle and role chip, with a provenance glyph (◦ user-created, ▸
+  agent-spawned) drawn from the session's on-disk metadata.
+- Counts: urgent-lane tally (merging `inbox/urgent/<handle>/` and
+  `inbox/urgent/all/`) in red, plus the unread count of
+  `inbox/direct/<handle>/`.
+- Last-heartbeat age as "30s ago" / "4m ago" with the full RFC3339
+  timestamp on hover.
+- Current task line (truncated to ~80 chars) from the presence file's
+  `current_task`.
+
+Rows sort urgent-first, then by unread desc, then by freshest heartbeat.
+Clicking a row raises that handle's twapp window via the same
+`launch_session` focus path the launcher uses. When the coordinator's
+session has a `colab_group`, the pane is scoped to that group; a
+coordinator without a group sees every active handle in the shared
+mailbox. Non-coordinator sessions never render the pane (single-session
+and worker UX are byte-unchanged). The backend lives in
+[`src-tauri/src/gui/fleet.rs`](src-tauri/src/gui/fleet.rs) and the React
+side in [`src/components/FleetPane.tsx`](src/components/FleetPane.tsx),
+both with unit tests.
+
 ### Spawning a worker agent
 
 twapp works well as a terminal wrapper for *interactive* sessions, but
--- a/docs/designs/agent-aware-ui.md
+++ b/docs/designs/agent-aware-ui.md
@@ -283,6 +283,17 @@ heartbeat loop.
 
 ### 3.3 Coordinator dashboard
 
+> **Partially landed** (ui-fleet-pane PR): the Fleet sidebar pane ships
+> read-only, reading `presence/<handle>.json` plus direct-inbox and
+> urgent-lane file counts via the new `list_fleet` Tauri command. Rows
+> are status-dot / handle / role / provenance / unread / urgent /
+> heartbeat-age; row-click raises the target session's window via the
+> existing `launch_session` focus path. Dashboard-mode expansion,
+> inbox/broadcast panes, the spawn timeline, and the quick-action
+> context menu remain separate PRs (§3.3 dashboard-mode, §3.5, §3.7).
+> Scoping: when the coordinator's session declares a `colab_group`, the
+> pane is scoped to that group; otherwise every handle is listed.
+
 Activates when the current session declares `role: coordinator` **and**
 at least one other session in the shared dir has a live
 `presence/<handle>.json`. Below that threshold, nothing changes in the
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -278,6 +278,13 @@ twapp msg presence list --stale
 twapp msg presence get worker-a
 ```
 
+If the coordinator is running inside a twapp window (role `coordinator`
+in `.twapp-session.json`), the same view is surfaced live in the
+sidebar's **Fleet** pane — one row per active handle with status dot,
+unread / urgent counts, and last-heartbeat age, refreshed every 5s.
+Click a row to raise that worker's twapp window. The CLI forms above
+remain the source of truth for scripting and terminals.
+
 Three liveness states the coordinator reasons about:
 
 - **processing / idle** — heartbeat within `5 × poll_interval_sec`. Alive.
--- a/src-tauri/src/gui/fleet.rs
+++ b/src-tauri/src/gui/fleet.rs
@@ -1,0 +1,466 @@
+//! `list_fleet` — coordinator fleet pane backend.
+//!
+//! Aggregates `<mailbox>/presence/<handle>.json`, inbox message counts, and
+//! session-directory metadata into one snapshot per active handle. Consumed by
+//! `src/components/FleetPane.tsx` on a 5s poll.
+//!
+//! Presence dormancy math lives in `cli::msg_presence::is_dormant`; this module
+//! only decorates those records with the per-handle inbox counts and with
+//! launcher-discovered metadata (provenance, colab_group, session directory)
+//! so row-clicks can raise the target window via `launch_session`.
+//!
+//! Scoping: when `colab_group` is `Some`, rows whose session file declares a
+//! different colab_group are excluded. Handles with no discoverable session
+//! file are kept (the presence record is authoritative for a handle's
+//! liveness; session-on-disk is a best-effort decoration).
+
+use serde::{Deserialize, Serialize};
+use std::collections::HashMap;
+use std::path::{Path, PathBuf};
+
+use crate::cli::msg::{direct_dir, inbox_dir, resolve_mailbox_dir, urgent_dir};
+use crate::cli::msg_presence::{is_dormant, list_presence, PresenceFile, PresenceStatus};
+use crate::cli::session::{list_sessions, SessionData};
+
+/// One row in the coordinator fleet pane.
+///
+/// The status string is the presence-recorded status when the heartbeat is
+/// fresh, but flips to `"dormant"` when `is_dormant` says so regardless of what
+/// the file says (§2.6 derives dormancy; the file can lag behind the truth).
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct FleetAgent {
+    pub handle: String,
+    /// `"processing" | "idle" | "dormant"` — derived, not the raw file field.
+    pub status: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub role: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub provenance: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub colab_group: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub current_task: Option<String>,
+    /// RFC3339 timestamp from the presence file (pass-through).
+    pub last_heartbeat: String,
+    /// Age of last heartbeat in seconds relative to the server's now. `None`
+    /// when the stamp was unparseable.
+    pub last_heartbeat_age_sec: Option<i64>,
+    pub poll_interval_sec: u64,
+    pub dormant: bool,
+    /// Count of direct messages sitting in `inbox/direct/<handle>/`. Cursor-
+    /// aware unread math is out of scope for this PR — it's a file count.
+    pub unread_count: u64,
+    /// Count of urgent-lane messages: `inbox/urgent/<handle>/` plus
+    /// `inbox/urgent/all/`. Blocker-priority is folded in (the urgent lane
+    /// already holds both urgent and blocker symlinks per PR-4).
+    pub urgent_count: u64,
+    /// Session working directory, discovered by matching `name == handle`
+    /// among `list_sessions(work_directory)`. `None` when the handle has no
+    /// on-disk session (e.g. a spawned agent that hasn't written session
+    /// metadata yet, or a handle that lives purely in the mailbox).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub directory: Option<String>,
+}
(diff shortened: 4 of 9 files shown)


piekstra (commented):
**Ship it.** Fleet pane implements design §3.3 read-only item (§5 roadmap PR-3) cleanly; pure `build_fleet_rows` is the right test seam; dormancy is derived correctly per §2.6; colab_group scoping handles the "during spawn" gap properly.

## Policy scan

Clean.

## What's right

### Pure builder separation

`build_fleet_rows` is a pure function taking `(presence, inbox_counter_closure, session_index, colab_group_filter, now)` → rows. No Tauri state, no filesystem access. The `list_fleet` Tauri command is a thin wrapper that resolves real state + defers to the pure builder. 4 of 7 tests exercise `build_fleet_rows` directly — the right seam.

### Dormancy correctness

`status` on the row is the **derived** state, not the file's. If `is_dormant` says the handle is stale, the row reads `"dormant"` regardless of what the presence file's own `status` claims. That matches §2.6's "dormancy is derived, not written" and the PR-5 contract I flagged on #61.

`last_heartbeat_age_sec` is computed only when `parse_from_rfc3339` succeeds, returning `None` on malformed timestamps. The sort comparator handles `None` explicitly (unparseable → sort last) — no panic on corrupt files.

### Colab group scoping — the right default

The "pass-through handles with no session metadata" behavior is the load-bearing decision here. The docstring names it:

> "filtering them out would mean a group=X coordinator loses visibility on any handle that hadn't written session metadata yet, which is the common case during spawn."

That's right. A handle that heartbeats before its `.twapp-session.json` is written is visible to the coordinator's group until the metadata catches up. If the session later declares a conflicting group, it disappears from view. Graceful.

### Counts

- `count_files_in` — regular files only, dotfiles skipped. Not recursive. Matches the flat-per-recipient layout from #60.
- `count_direct` = `inbox/direct/<handle>/*` count. Un-cursor'd file count (not unread-math); PR body calls this out as scope.
- `count_urgent` sums both `inbox/urgent/<handle>/` and `inbox/urgent/all/` — the blocker-priority-folds-into-urgent convention from #53/#57.

### Sort

`urgent DESC > unread DESC > heartbeat-age ASC (freshest first) > handle ASC` matches design §3.3 verbatim. `test_sorts_urgent_then_unread_then_freshness` pins the ordering across all four tiebreaks.

### Session index

Reuses `list_sessions(global_config.work_directory)` — same helper the launcher uses, so the Fleet pane and the launcher stay in lockstep on what a session is. `handle → (dir, SessionData)` built once per `list_fleet` call. `config::GlobalConfig::load()` error → empty index (still render the pane from presence alone, just no directory decoration). Sensible.

### Gating

Frontend-side: the `FleetPane` is rendered only when `role == "coordinator"`. Non-coordinator sessions (including single-session users) never render the pane. Design §3.6 preservation. ✓

### Test coverage (7 tests)

- `build_rows_passes_through_presence_and_inbox_counts` — happy path.
- `build_rows_marks_dormant_when_heartbeat_stale` — derived dormancy overrides file status.
- `build_rows_filters_by_colab_group_when_set` — scope filter + the "no metadata = pass through" carve-out.
- `build_rows_sorts_urgent_then_unread_then_freshness` — full sort contract.
- `count_files_in_counts_only_regular_files` — dotfile + subdir skip.
- `count_urgent_sums_handle_dir_and_all_dir` — merge correctness.
- `end_to_end_with_real_presence_dir` — integration test with `write_presence` + disk round-trip.

## Non-blocking nits

- **`unread_count` is a file count, not cursor-aware unread math.** A coordinator who has `fetch --mark-read`'d all direct messages will still see them counted as unread here. PR body names this as scope — OK for the first version of the pane; a follow-up can consult `cursors/<handle>.jsonl` to subtract already-read. Worth a tooltip on the pane saying "count of files in inbox/direct/<handle>/" so the coordinator knows what they're looking at.
- **5-second poll cadence** is tight but fits the "fleet view refresh" intent. With 20 presence files and 20 inbox dirs, each poll does ~40 `read_dir` calls — cheap, but worth watching if the fleet grows past that. Non-blocking today.
- **`list_fleet` reads the session index on every poll** (no cache). 5s × N sessions scans the work_directory repeatedly. For N < 50 sessions this is negligible; scale concern only. Non-blocking.
- **CI still `IN_PROGRESS`** at the time of this review. Structure is correct; expecting green. If CI fails, the most likely culprit is the frontend test file (`FleetPane.test.ts`) which I haven't read in full — its 86 lines are unit tests for the component, not the backend.

### Design doc

§3.3 gets a "Partially landed" callout naming the shipped scope (read-only pane; dashboard-mode expansion, inbox/broadcast panes, spawn timeline, context menu deferred). Honest about what ships vs what's aspirational — same pattern as #59's §3.1/§3.2 callouts.

## Summary

Right-sized read-only Fleet pane; pure builder with dense test coverage; dormancy derived correctly; colab_group scoping graceful during spawn; reuses existing `list_sessions` / `is_dormant` / `count_files_in`-style plumbing so it's consistent with the messaging + launcher layers. Ship it.

(Note: #62 is the parallel context-menu PR and explicitly ships the `AgentContextMenu` component in isolation; a follow-up PR to this one — or the merged result — wires it into FleetPane's rows.)

piekstra (commented):
**Ship it (COMMENTED / second-pair-of-eyes from `twapp-reviewer`).** Thanks for the direct ping — caught this right after course-correcting my poll cadence.

## Policy scan
Clean. No private-project terms. "fleet", "presence", "dormant", "heartbeat" are messaging-design vocabulary.

## Correctness

### `build_fleet_rows` — the pure builder
- Signature takes `presence`, `count_inbox_for` closure, `session_index`, `colab_group_filter`, `now`. That's the right testing seam — `build_rows_passes_through_presence_and_inbox_counts` injects fake counts via the closure, no mailbox fixture required.
- Dormancy derivation is load-bearing: the builder flips status to `"dormant"` whenever `is_dormant(&p, now)` says so, regardless of what the file field claims. `build_rows_marks_dormant_when_heartbeat_stale` pins both invariants (`rows[0].dormant == true` AND `rows[0].status == "dormant"`) — the on-file status lagging behind truth is a real PR-5 scenario, so this is the right place to guard it.
- Sort order `urgent desc → unread desc → heartbeat age asc → handle name tiebreak` matches §3.3 exactly. `build_rows_sorts_urgent_then_unread_then_freshness` locks all three tiers with a 3-agent fixture that intentionally has each lane win a different agent. Good.
- `last_heartbeat_age_sec: None` agents sort *after* `Some` agents via the explicit Ordering match — that's the right default (a handle with a garbage timestamp shouldn't jump to the top of the list). Subtle enough to warrant a test someday; not blocker.

### Group filter — "handles with no session file are kept"
The comment in the code is the important documentation here:
> `ghost` has no session file → passes through the filter (see comment in build_fleet_rows for the reasoning).
And the reasoning is right: during spawn, a worker has a presence heartbeat before it has written session metadata. If the filter dropped unknown-group handles, a group=X coordinator would briefly lose visibility on workers it JUST spawned — exactly when visibility matters most. `build_rows_filters_by_colab_group_when_set` exercises the three-way split (grp-a kept, grp-b dropped, unknown-group `ghost` kept). Correct disposition.

### Inbox counting
- `count_files_in` skips dotfiles (`.DS_Store` et al) + subdirectories + missing dirs → 0. `count_files_in_counts_only_regular_files` covers all three.
- `count_urgent` sums `urgent/<handle>/` + `urgent/all/`. Matches PR-4's symlink-lane semantics (blockers land in both the per-handle dir AND in `all/` for broadcasts). `count_urgent_sums_handle_dir_and_all_dir` asserts the specific "beta has no handle dir but still sees urgent/all/" case. Good.
- Not recursive — correct for the flat `inbox/direct/<handle>/` and `inbox/urgent/<handle>/` layouts from PR-3 / PR-4. If PR-6 introduces channel-scoped sub-hierarchies, this will need to evolve.

### Tauri command surface
- `list_fleet` is best-effort on `GlobalConfig::load()` — if global config is unloadable, session decoration is empty but rows still render from presence alone. "brand-new install" case handled. Good.
- `chrono::Utc::now()` passed in once per call; the builder is deterministic for a given `now`. `end_to_end_with_real_presence_dir` test exercises the full write_presence → list_presence → build_fleet_rows round-trip without mocking between layers — nicely catches any type drift.

### Frontend — `FleetPane.tsx`
- Component gated on `isCoordinator` — non-coordinator sessions get `null` and no chrome. Workers see no new UI. §3.6 single-session preservation honored.
- Poll interval 5s. Only runs when mounted + `isCoordinator === true` (the effect early-returns otherwise). Matches PR-57's pattern but *without* UrgentInbox's auto-collapse-on-empty — correct design call here: coordinators care about the pane being present even when the fleet is quiet, so collapsing would be worse UX than a "No workers online" line.
- `fetchSeqRef` drop-stale-resolves — same pattern as UrgentInbox. Consistent, catches the two-in-flight race.
- `tauriOpener`: `invoke("launch_session", { sessionId: "", directory: agent.directory })`. This assumes `launch_session` treats empty `sessionId` as "find-by-directory and focus". I haven't read that code path but the PR body + briefing both describe it as existing focus-if-running behavior via `open -a <instance>.app`. Worth a one-line test pin (or a code comment at the call site) so a future refactor of `launch_session` doesn't silently break the row-click flow.
- `handleRowClick` no-ops when `a.directory` is missing, with a title attr explaining "no on-disk session yet". Title is also shown on `<li role={a.directory ? "button" : undefined}>` so screen readers know which rows are clickable. Nice accessibility touch.
- `statusDotClass` has a "guard the UI against a hand-crafted response that left status=processing with dormant=true" override. That's belt-and-braces — the Rust builder already flips, but the TS defense is cheap and documented. Good.

### Props injection
- `fetcher` and `opener` are both overridable in tests. `pollMs` too. This is the clean pattern — `UrgentInbox` did the same thing and it's paying off now that multiple components use it.

## Scope discipline
Files match the PR body: new `gui/fleet.rs` + `FleetPane.{tsx,test.ts}` + CSS + App.tsx wiring + docs. 1093 additions, 0 deletions — genuinely additive, no drift into §3.5 (quick-action context menu — PR #62's territory), §3.3 dashboard-mode, or §3.7 spawn timeline.

Note: PR #62 also touches `src-tauri/src/gui/mod.rs` (adding `agent_actions::*` to the handler list) and `src/App.css` (adding `.agent-menu-*` styles). Those changes are additive and non-overlapping, but **merge order matters**: whichever lands second will need a trivial rebase to add its module line to the alphabetized handler list. Both PRs have CLEAN merge state today — first-to-merge wins.

## Tests
- Rust: 7 new (pass-through, dormant flip, colab_group filter, sort order, count_files, count_urgent, end-to-end with real presence). 239/239 pass per PR body.
- TS: 9 new (`formatAge` 4 branches including NaN + undefined + negative clamp, `summarize` counts, `statusDotClass` routing including dormant override, `truncateTask` truncation + edge cases). 136/136 pass per PR body.

Gaps (not blockers):
- No test exercises the 5s poll interval itself — a vitest with fake timers + an injected `fetcher` tracker would verify that the component actually calls `fetcher` per tick. The fetch invariant would silently regress if someone broke the effect dependency array.
- No test for the row-click opener call — an injected `opener` mock asserting it's called with `a.directory` on click would cover the UI → Tauri handoff.
- No test for the "sort with `None` heartbeat age" edge case — the Ordering match has three branches, only one (`Some/Some`) is exercised by the current fixtures.

## Docs consistency
- README "Coordinator fleet pane" subsection is right-sized: explains the status-dot mapping, the heartbeat-age formatter, the urgent+unread counts, the sort order, the colab_group scoping, and the non-coordinator default. Operators know what to expect.
- `docs/designs/agent-aware-ui.md` §3.3 gets a "Partially landed" callout at the top that correctly scopes what shipped vs what didn't (dashboard mode, inbox panes, spawn timeline, quick-action menu). Good hygiene — the design doc is the coordination source of truth for the three parallel wave-3 PRs.
- `skills/agent-coordinator/SKILL.md` §5 "Coordinator dashboard" paragraph correctly points at both the CLI (`twapp msg presence list`) as source of truth AND the new UI pane as the live surface. That's the right positioning — CLI stays scriptable.

- `list_fleet` doesn't surface a reason when `GlobalConfig::load()` fails — a `log::warn!` on that path would help a confused operator diagnose "fleet pane shows handles without directories." The pane still works, but the missing directory on every row would otherwise be puzzling.
- `build_fleet_rows`'s handle-name tiebreak happens inside the sort — it's case-sensitive (`a.handle.cmp(&b.handle)`), which is fine today since handles are lowercase-dashes by convention. If a user ever mixes cases, sort could surprise. `a.handle.to_lowercase().cmp(&b.handle.to_lowercase())` is cheaper than you'd think and future-proofs.
- `FleetPane` has no loading state visible to the user — an empty panel while the first fetch is in-flight. If fetchers are slow, that's momentarily confusing. `{!loaded && agents.length === 0 && <div>Loading fleet…</div>}` would close that.
- The provenance glyph ▸ is documented in the README as "agent-spawned" — consider a legend or tooltip-on-hover for the chevron glyph. `aria-label="spawned"` is already there; sighted users may still find it cryptic.
- Dependency note in the PR body is accurate: "presence files are only present when an agent manually runs `twapp msg presence heartbeat`, so the pane legitimately shows 'No workers online' in most repos today." This is the exact operator-facing friction that will drive demand for the heartbeat-inside-/loop follow-up. No action needed on this PR.

Clean, well-tested, well-scoped wave-3 pane. Pure builder + decoration pattern is the right shape. Ready to merge once manual smoke happens (blocked only on an active coordinator fleet being available — reasonable to defer). First-to-merge between #62 and #63 just needs the trivial alphabetical rebase on the other.

PR #64, opened 2026-04-21
feat(ui): coordinator spawn/teardown timeline

piekstra:
## Summary

- Adds a **Timeline** panel to the coordinator dashboard (below the fleet pane from PR #62). Lists spawn / claim / reclaim / release / offboard / dead events for the co-lab, newest-first, with a 7-day window + *Load more* pagination and a handle-substring filter. Renders only when `sessionRole === "coordinator"`, so single-session and worker UX are byte-unchanged.
- Backend `list_timeline_events` (`src-tauri/src/gui/timeline.rs`) aggregates four streams:
  - **Spawn** — `SessionData.created` per `.twapp-session.json` under the work directory (respects `colab_group`).
  - **Claim / reclaim / release** — `<mailbox>/claims/<lane-id>/owner.json` + `released.json`, plus the `.released-<ts>/` archives. Reuses `ClaimOwner` / `ClaimRelease` from the lane-claim primitive.
  - **Offboard** — `to: [all]` broadcasts whose subject or body contains "offboard" (case-insensitive), walked across `inbox/broadcast/` and `archive/` (flat + dated subdirs from PR #51's rotation).
  - **Dead** — presence handles whose `last_heartbeat` is older than `15 × poll_interval_sec` (3× the dormant threshold, to drain noise from briefly-paused agents).
- Filenames in the broadcast archive use the compact `YYYYMMDDTHHMMSSZ` shape, so the scanner normalizes to RFC3339 before comparing against the `since` cutoff.
- 11 Rust tests (claim/release pairs, stale-reclaim, archived `.released-*` dirs, offboard subject+body match, 7-day cutoff, dead threshold, unparseable timestamps, window/sort/filter/pagination).
- 8 frontend tests (chip label, sort-newest-first no-mutate, case-insensitive handle filter, page-merge dedup that keeps different-kind rows under the same handle).
- Matches §3.3 of `docs/designs/agent-aware-ui.md` (now updated to mark spawn-history as landed) and re-uses the same claim/release primitive documented in `docs/designs/worker-coordination.md`.

The original §3.3 sketched a horizontal hover-strip; this PR ships a chronological list in the sidebar column instead — fits the existing panel vocabulary (fleet pane rows, urgent-inbox rows) better, and the timeline-chip kinds map cleanly to a kind-specific color via `.timeline-chip-{spawn,claim,reclaim,release,offboard,dead}`.

## Test plan

- [x] `cargo test --lib` — 262 passed (11 new in `gui::timeline::tests`)
- [x] `npm test` — 157 passed (8 new in `src/components/TimelinePane.test.ts`)
- [x] `npm run dev` smoke (Vite serves 200 at :1420, no import errors)
- [ ] Visual verification — the panel is gated on `sessionRole === "coordinator"`, so it only renders when the current session's `.twapp-session.json` has that role. Reviewer: open a coordinator session with a populated mailbox (presence + claims + a "offboard" broadcast) to exercise rows of each kind; confirm the *Load more* button and the handle-filter input behave as described.

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,6 +474,31 @@ and worker UX are byte-unchanged). The backend lives in
 side in [`src/components/FleetPane.tsx`](src/components/FleetPane.tsx),
 both with unit tests.
 
+### Spawn / teardown timeline
+
+Directly below the fleet pane, coordinator sessions see a **Timeline**
+panel that lists spawn, claim, reclaim, release, offboard, and dead
+events for the co-lab, newest-first. Polls the `list_timeline_events`
+Tauri command every 30 seconds and defaults to the last 7 days; a
+*Load more* button extends the window 7 days further back per click.
+A free-text filter narrows rows to handles matching a substring.
+
+Events are aggregated from:
+
+- **Spawn** — `SessionData.created` for each `.twapp-session.json`
+  under the global work directory (scoped by `colab_group` when set).
+- **Claim / reclaim / release** — `<mailbox>/claims/<lane-id>/owner.json`
+  and `released.json`, plus the `.released-<ts>/` archived dirs from
+  repeat claims. Structural records from the lane-claim primitive
+  (§1.4 of [`worker-coordination.md`](docs/designs/worker-coordination.md)).
+- **Offboard** — `to: [all]` broadcasts whose subject or body contains
+  "offboard" (case-insensitive).
+- **Dead** — presence handles whose `last_heartbeat` is older than
+  `15 × poll_interval_sec` (3× the dormant threshold).
+
+Backend: [`src-tauri/src/gui/timeline.rs`](src-tauri/src/gui/timeline.rs).
+Frontend: [`src/components/TimelinePane.tsx`](src/components/TimelinePane.tsx).
+
 ### Spawning a worker agent
 
 twapp works well as a terminal wrapper for *interactive* sessions, but
--- a/docs/designs/agent-aware-ui.md
+++ b/docs/designs/agent-aware-ui.md
@@ -288,11 +288,21 @@ heartbeat loop.
 > urgent-lane file counts via the new `list_fleet` Tauri command. Rows
 > are status-dot / handle / role / provenance / unread / urgent /
 > heartbeat-age; row-click raises the target session's window via the
-> existing `launch_session` focus path. Dashboard-mode expansion,
-> inbox/broadcast panes, the spawn timeline, and the quick-action
-> context menu remain separate PRs (§3.3 dashboard-mode, §3.5, §3.7).
-> Scoping: when the coordinator's session declares a `colab_group`, the
-> pane is scoped to that group; otherwise every handle is listed.
+> existing `launch_session` focus path. Dashboard-mode expansion and
+> the inbox/broadcast panes remain separate PRs (§3.3 dashboard-mode,
+> §3.5). Scoping: when the coordinator's session declares a
+> `colab_group`, the pane is scoped to that group; otherwise every
+> handle is listed.
+>
+> **Additionally landed** (ui-spawn-timeline PR): the spawn/teardown
+> timeline strip is now a sibling panel under the fleet pane. Backend
+> `list_timeline_events` aggregates spawn events (session-metadata
+> `created`), claim/reclaim/release events (from `<mailbox>/claims/`
+> and its `.released-*` archives), offboard broadcasts (`to: [all]`
+> messages whose subject or body contains "offboard"), and dead events
+> (presence heartbeat older than 15× `poll_interval_sec`). Frontend
+> polls every 30s, filters by handle, and paginates older events by
+> walking `beforeTs` backward 7 days per click.
 
 Activates when the current session declares `role: coordinator` **and**
 at least one other session in the shared dir has a live
@@ -365,9 +375,17 @@ the terminal is the right shape.
   not a direct recipient. This is the only cross-agent peek the
   dashboard does, because "a blocker anywhere in the fleet" is load-
   bearing for the coordinator.
-- **Spawn timeline** — horizontal strip below the panes, one tick per
-  spawn / offboard event in the last 2 hours. Hover = briefing title,
-  exit reason, duration. See §3.7 on the event-log dependency.
+- **Spawn timeline** — chronological list below the fleet pane
+  (ui-spawn-timeline PR). One row per spawn / claim / reclaim / release
+  / offboard / dead event, newest first, default window of 7 days with
+  a *Load more* paginator that extends 7 days back per click. Each row
+  shows a ts chip, a kind chip (styled per kind via
+  `.timeline-chip-{spawn,claim,reclaim,release,offboard,dead}`), the
+  handle, and a one-line description. A free-text handle filter
+  narrows the list without re-fetching. The originally-sketched
+  horizontal strip with hover tooltips is deferred — the list shape
+  fits the sidebar-column layout more naturally and lines up with the
+  fleet pane's vocabulary.
 
 In sidebar-only (non-expanded) mode, only the fleet pane summary
 (`6 active · 1 urgent`) is visible, and it deep-links to the full
--- a/src-tauri/src/gui/mod.rs
+++ b/src-tauri/src/gui/mod.rs
@@ -11,6 +11,7 @@ pub mod pty;
 pub mod sessions;
 pub mod shell_env;
 pub mod tickets;
+pub mod timeline;
 pub mod title;
 pub mod types;
 
@@ -130,6 +131,7 @@ pub fn run(args: GuiArgs) {
             msg::send_message,
             msg::fetch_messages,
             fleet::list_fleet,
+            timeline::list_timeline_events,
             agent_actions::focus_agent_window,
             agent_actions::stop_agent,
             agent_actions::list_agent_prs,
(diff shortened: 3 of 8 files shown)


piekstra (commented):

## Policy scan
Clean. No private-project terms. "spawn", "teardown", "offboard", "dead", "claim" are all established vocabulary.

## Correctness

### Four event streams — each one right-shaped

**Spawn (`scan_spawn_events`)**
- Uses `SessionData.created` as the authoritative spawn timestamp. The doc-comment calls out *why*: presence files get overwritten in place per §2.6, so presence first-write is not a stable spawn record. Correct.
- `colab_group_filter` passes through unknown-group sessions (matching the fleet pane's reasoning). That consistency is load-bearing — a `grp-X` coordinator would otherwise lose sight of a session it JUST spawned before the group field hit disk.
- Description varies with role + provenance presence: `spawned — role {role} · {prov}` / `spawned — role {role}` / `spawned — {prov}` / `spawned`. Readable, doesn't crash on missing metadata.

**Claim / Reclaim / Release (`scan_claim_events`)**
- Reuses `ClaimOwner` / `ClaimRelease` from the lane-claim primitive. Zero duplication of parse logic.
- `name.split(".released-").next()` to recover the original lane id from archived dirs. Relies on `validate_lane_id` in `msg_claim` rejecting `.` in lane ids — worth noting that's a load-bearing invariant (comment already does). If someone ever loosens lane-id validation, this splitter silently truncates. `claim_scanner_reads_released_archive_dirs` locks the behavior for `PR-42.released-20260420T120000Z`.
- `reclaimed_from` presence → reclaim event; absence → plain claim. `claim_scanner_marks_reclaim_when_field_present` covers this.
- A live lane can emit up to three events (claim/reclaim + release when released.json also present). `claim_scanner_emits_claim_and_release_for_live_lane` covers the happy path.

**Offboard (`scan_offboard_events`)**
- Walks `inbox/broadcast/` + `archive/` recursively (`walk_md_files`), skips symlinks to avoid double-counting through PR-3's split + PR-4 urgent lane + PR-7's date-rotated archive. `canonicalize`-based `seen` set drops canonical-path duplicates.
- Filter: `to contains "all"` + subject OR body contains "offboard" (case-insensitive). `offboard_scanner_matches_subject_or_body` covers both axes with a `worker-c` negative control.
- `since` cutoff applied *inside* the scanner (not left to `assemble_events`), which is a nice perf win — don't canonicalize year-old files for nothing. `offboard_scanner_skips_messages_older_than_since` verifies.
- `normalize_compact_ts` handles both the compact `YYYYMMDDTHHMMSSZ` filename shape and already-RFC3339 frontmatter `ts`. The tests pin the conversion + the length-check rejection of malformed input.

**Dead (`scan_dead_events`)**
- Threshold `DEAD_MULTIPLIER = 15` → `15 × poll_interval_sec`, documented as 3× the dormant multiplier (§2.6). Rationale: "dead means two additional poll cycles past dormant onset," which drains noise from briefly-paused agents. Good justification, good magic-number hygiene.
- Event `ts` is `last_heartbeat` (the OLD timestamp), not `now()` — so the event stays stable across scans. Comment explicitly calls this out.
- Unparseable heartbeat → still emit a `dead` event, tagged with `age_sec: 0`, ts = `now()` (stretched the "use the heartbeat ts" rule, but alternative is dropping the data). `dead_scanner_handles_unparseable_timestamp` pins this choice.

### `assemble_events` — pure builder
- Signature takes all four streams + window/pagination/filter/limit params. Pure, testable, doesn't touch the filesystem. `assemble_sorts_newest_first_and_respects_window`, `assemble_handle_filter_matches_substring_case_insensitive`, `assemble_before_ts_paginates` each exercise one axis cleanly.
- Sort: `b.ts.cmp(&a.ts).then_with(|| a.handle.cmp(&b.handle))` — newest-first, alphabetical handle tiebreak for stable ordering. No sort-randomness between equal-ts events.
- Handle filter applied lower-cased → case-insensitive substring. No regex-engine pathological inputs.
- `limit` truncation after sort. Works correctly when limit > total (no-op).

### Tauri command (`list_timeline_events`)
- `since_ts` default: 7 days ago. `before_ts` optional for pagination.
- Best-effort `GlobalConfig::load()`: fails silently to empty spawns list. Matches the fleet pane's pattern — pane still renders from mailbox-only streams. Good.

### Frontend (`TimelinePane.tsx`)
- `isCoordinator` gate → non-coordinators get `null`. §3.6 preserved.
- Poll 30s (less aggressive than fleet pane's 5s — appropriate for a mostly-historical view).
- `doRefresh` REPLACES events, `doLoadMore` MERGES pages via `mergeEvents` which dedups by `(ts, handle, kind, description)` tuple. The description-includes-kind-id tiebreak keeps "a's claim on PR-1" and "a's reclaim on PR-1" as separate rows, which is right. `keeps two events with the same handle but different descriptions` pins this.
- `fetchSeqRef` stale-drop — same pattern as `UrgentInbox` and `FleetPane`. Consistent.
- `reachedEnd` gate: when a load-more returns an empty page, disable the button + show "No more events". Right UX — better than infinite "Load more" on an exhausted mailbox.

## Scope discipline
Files match the PR body: new `gui/timeline.rs` + `TimelinePane.{tsx,test.ts}` + CSS + App.tsx wiring + docs. 1585 additions, 8 deletions (the §3.3 callout rewrite). No drift into fleet-pane territory, no overlap with PR #62's context-menu territory.

Merge state: CLEAN today, CI green. Files it touches in `gui/mod.rs` (1 module + 1 handler entry) are additive, non-overlapping with any other open PR I can see.

## Tests
- Rust: 11 (claim pair + reclaim + archive dirs / offboard subject-or-body + since-cutoff / dead threshold + unparseable-ts / assembly sort + handle filter + before-ts / normalize_compact_ts). 262/262 pass per PR body.
- TS: 8 (`chipLabel` known + unknown passthrough / `sortNewestFirst` sort + no-mutate / `filterByHandle` empty + case-insensitive / `mergeEvents` dedup + different-kind-same-handle preservation). 157/157 pass per PR body.

Gaps (not blockers):
- **Dead-description age drift**: `TimelineEvent::dead`'s `description` includes `age_sec`, which grows between polls. Two scans 30s apart produce the same `(ts, handle, kind)` but *different* description. Today this doesn't matter because `doRefresh` REPLACES events (fresh list, fresh description). But `doLoadMore`'s `mergeEvents` would see them as distinct on the exact boundary case where the same dead handle shows up in both the current window AND an older page. Low-prob in practice. Fix options: drop `age_sec` from description (recompute in UI from `ts`), or change dedup key to `(ts, handle, kind)` only.
- **Offboard false-positives**: a broadcast *discussing* the word "offboard" (e.g., design-doc update) would be flagged as an offboard event. The PR body acknowledges this ("free-form — the design doesn't prescribe a subject yet"). Worth a follow-up convention ("broadcasts with `re: offboard —` subject only") documented in the skill doc.
- **`normalize_compact_ts` relaxed mode**: `ts.contains('-') || ts.contains(':')` as the "already RFC3339" heuristic would accept a malformed string like `-foo-bar-`. Not a real-world concern (mailbox timestamps are machine-written) but a strict regex would be cheaper-to-trust.
- No test for the `sort_by ... handle.cmp` tiebreak — the `assemble_sorts_newest_first_and_respects_window` fixture uses distinct `ts` values so tiebreak never fires. One equal-ts case would lock it.
- No test for the 30s poll ticking — same gap I flagged on #63. Injectable `pollMs` is there (good); a vitest with fake timers would close it.

## Docs consistency
- README "Spawn / teardown timeline" subsection sits right after the fleet pane subsection. Lists the four event streams with their sources, the 7-day + load-more pagination, the handle filter, the 30s poll. Right placement.
- `docs/designs/agent-aware-ui.md` §3.3 gets a "Additionally landed" callout and §3.3 (spawn timeline bullet) gets updated from the horizontal-strip sketch to the chronological-list shape that actually shipped. The justification ("fits the sidebar-column layout more naturally, lines up with the fleet pane's vocabulary") is worth keeping — explains why the PR diverged from the original design.

- The dead-event description growth is the one I'd most like to see addressed before ship (it's not a bug today but it's a latent one). Dropping `age_sec` from description and letting the UI compute it from `ts` (same `formatAge` pattern as the fleet pane) would also shrink Tauri payload size on repeat polls.
- `walk_md_files` doesn't scope to `broadcast/` subdirs before reading files — it walks ALL `archive/*` files and filters by `to: [all]` at parse time. For a long-lived mailbox with a big `archive/direct/*` tree, the offboard scan parses every direct message only to discard it. A one-line "skip `archive/**/direct/` / `archive/**/urgent/` / `archive/**/cursors/`" would cut that cost.
- `scan_offboard_events` doesn't cache parsed messages, so if you later add another broadcast-based scanner (e.g. "resumed" events), you'd re-walk the same tree. Borderline — today's cost is fine.
- `chipLabel` is a full switch statement that hard-codes the six kinds. If kinds grow, consider a `KIND_LABELS: Record<string, string>` lookup so the TS type and the label table live together.
- CSS `.timeline-chip-*` colors use mid-weight hex + 55% border alpha consistently across six kinds. Good visual coherence; `reclaim`'s amber is distinguishable from `spawn`'s green and `offboard`'s purple. No accessibility concern on contrast.

Solid PR. The four-stream aggregation shape is exactly right, the pure `assemble_events` builder makes the window/sort/filter/pagination math testable without fixtures, and the `DEAD_MULTIPLIER` rationale is the kind of magic-number comment that saves a future reader 20 minutes. Ready to merge once manual smoke-test can exercise a populated coordinator fleet (the usual "deferred pending a live coordinator" caveat from wave-3 siblings).

PR #65, opened 2026-04-21
feat(msg): channels — design PR-6

piekstra:
## Summary

PR-6 of [`docs/designs/agent-messaging.md`](docs/designs/agent-messaging.md) §2.3. Channels are topic-scoped fan-in — a sender writes to a *name*, not a handle, and any subscriber watching that name picks up the traffic. Replaces the ad-hoc `reviewer-standby`-style pseudo-handle hack that previously leaked channel state into the receiver slot.

The addressing primitive (`to: channel:<name>` → `inbox/channel/<name>/`) already landed in PR-3 — PR-6 fills in the observability lookups the design spec calls for and fixes the edge cases that blocked channel use end-to-end.

### What ships

- **`twapp msg send --channel <name>` form**: the positional `<to>` is now optional; the one positional is reinterpreted as the body when `--channel` supplies the routing. Positional `channel:<name>` form already worked and is unchanged.
- **`twapp msg fetch --channel <name> [--for <handle>]`**: both flags can coexist — channel becomes the primary addressing check, and `--for` downgrades to "exclude messages I sent". Previously the stacked `matches_for_handle` filter dropped every channel message.
- **`twapp msg fetch --priority urgent --channel <name>`**: disables the urgent-lane fast path when `--channel` is set, since channel messages don't get `urgent/` symlinks today (deferred to a later PR).
- **New `twapp msg channel list [--format json]`** — every channel under `inbox/channel/` with a message count, empty channels included.
- **New `twapp msg channel subscribers <name> [--format json]`** — handles whose `presence/<handle>.json` `claims` array contains `channel:<name>`.

Subscription remains by-convention (design §2.3): senders don't consult the subscribers list; every subscriber scans its own claimed channels in its `/loop` fetch cycle. **Out of scope**: enforcement of subscription, per-channel ACLs, urgent-lane for channels.

### Test isolation cleanup

Adding `msg_channel`'s own `env_lock()` static tipped over a pre-existing flake: every `msg_*` test module had its own `OnceLock<Mutex<()>>`, so they didn't coordinate across modules even though they all mutate process-wide `TWAPP_MAILBOX_DIR`. Consolidated into a single `cli::test_env::lock()` shared by every test that touches env. The whole 263-test suite now passes in parallel (previously did only serially).

### Docs

- README "Channels (PR-6)" subsection + PR-6 added to the design coverage list.
- `skills/agent-coordinator/SKILL.md`: channels supersede the `-standby` handle hack; added a Channels subsection and updated the reviewer archetype row.
- `skills/spawn-agent/SKILL.md`: Example B (reviewer) shows the channel subscription pattern via `presence heartbeat --claims`.

## Test plan

- [x] `cargo build` — clean
- [x] `cargo test --lib` — 263 passed (was 251 pre-PR; added 12 channel + cmd_send tests)
- [x] `cargo clippy --lib` — no new warnings from this PR
- [x] E2E CLI smoke: send (both forms), broadcast --channel, channel list (pretty + json), channel subscribers (with and without presence seeds), fetch --channel, fetch --channel --for, unknown channel fetch
- [ ] Reviewer to verify the `fetch --for X --channel Y` semantics match their mental model — the PR interprets `--for` as "exclude self-sends" when `--channel` is also set, so a subscriber's loop doesn't re-surface its own messages.

### DIFF
--- a/README.md
+++ b/README.md
@@ -606,13 +606,12 @@ back. See
 [`docs/designs/agent-messaging.md`](docs/designs/agent-messaging.md) for
 the full model. Today's CLI covers **PR-1** (scaffolding), **PR-2**
 (threading), **PR-3** (directory split + read cursors), **PR-4**
-(priority lane), **PR-5** (presence / heartbeat), and **PR-7** (archive
-rotation) of that design. Sends land under a per-recipient layout
-(`inbox/broadcast/`, `inbox/direct/<handle>/`, `inbox/channel/<name>/`);
-every send also drops a grace-period symlink at the flat
-`inbox/<filename>.md` path so un-upgraded readers doing `ls inbox/`
-keep working. First-class channel subscriptions still land in a later
-PR.
+(priority lane), **PR-5** (presence / heartbeat), **PR-6** (channels),
+and **PR-7** (archive rotation) of that design. Sends land under a
+per-recipient layout (`inbox/broadcast/`, `inbox/direct/<handle>/`,
+`inbox/channel/<name>/`); every send also drops a grace-period symlink
+at the flat `inbox/<filename>.md` path so un-upgraded readers doing
+`ls inbox/` keep working.
 
 #### Configure the mailbox
 
@@ -847,6 +846,56 @@ Agents are expected to heartbeat on their /loop cadence (today, manually;
 a follow-up PR will wire heartbeats into the /loop skill itself). The
 `spawn-agent` and `agent-coordinator` skill docs call this out in the
 worker protocol.
+
+#### Channels (PR-6)
+
+Channels are topic-scoped fan-in: a sender writes to a *name*, not a
+handle, and any subscriber watching that name picks up the traffic.
+They replace the ad-hoc `-standby`-handle hack the pre-PR-6 corpus used
+to signal "a reviewer, currently in standby mode, please pick this up
+if that's you". Design §2.3.
+
+```bash
+# Send — either `channel:<name>` as the positional recipient, or
+# `--channel <name>`. Both forms write to inbox/channel/<name>/.
+twapp msg send channel:reviewers "anyone free to look at PR-91?"
+twapp msg send --channel reviewers "anyone free to look at PR-91?"
+
+# Multi-destination: direct + channel in the same send.
+twapp msg send worker-a --channel announcements --priority urgent \
+  "merge freeze starts at 5pm"
+
+# Broadcasts can also fan into a channel instead of every handle.
+twapp msg broadcast --channel reviewers-standby "PR-91 needs eyes"
+
+# Fetch — scope to a single channel. Optional --for filters out the
+# caller's own sends (so your loop doesn't re-surface messages you just
+# wrote).
+twapp msg fetch --channel reviewers
+twapp msg fetch --channel reviewers --for reviewer-a
+
+# Observability — what channels exist and who's listening.
+twapp msg channel list                      # pretty table of name + count
+twapp msg channel list --format json | jq
+twapp msg channel subscribers reviewers     # handles whose presence.claims holds channel:reviewers
+twapp msg channel subscribers reviewers --format json
+```
+
+Subscription is by-convention (design §2.3): a worker declares its
+interest by putting `channel:<name>` into its `presence/<handle>.json`
+`claims` array — typically via `twapp msg presence heartbeat --claims
+channel:reviewers,channel:announcements`. The claims list is coordinator-
+facing observability; senders don't consult it, and nothing enforces
+that a non-subscriber won't read the channel. Actual delivery = every
+subscriber scanning its own claimed channels in its /loop fetch cycle.
+
+> Unknown channel fetch is a no-op: `twapp msg fetch --channel
+> never-created-yet` returns `(no messages)` and exit 0, not a crash.
+
+Channel messages archive under the same daily rotation as every other
+message — drop a channel file into `archive/` (flat) and
+`twapp msg archive rotate` groups it under `archive/<YYYY-MM-DD>/`
+by its frontmatter `ts`, same as broadcasts and direct messages.
 
 #### Migrating the inbox layout
 
--- a/skills/agent-coordinator/SKILL.md
+++ b/skills/agent-coordinator/SKILL.md
@@ -311,7 +311,47 @@ so its workers can see it's alive.
 
 - `to: all` — broadcast (every online agent reads and archives).
 - `to: <handle>` — directed (only that agent archives; others ignore).
+- `to: channel:<name>` — topic-scoped fan-in; any subscriber reading that channel picks it up (see below).
 - `cc: <handle>` — optional courtesy copy.
+
+### Channels (PR-6) — topic-scoped fan-in
+
+Channels replace the `reviewer-standby`-style pseudo-handle hack the
+pre-PR-6 corpus used to signal "any reviewer, please pick this up".
+Instead of renaming the recipient slot, send to a channel name and let
+any worker that subscribes pick it up:
+
+```bash
+# Send — either positional `channel:<name>` or `--channel <name>`.
+twapp msg send channel:reviewers "PR-91 needs eyes"
+twapp msg broadcast --channel reviewers-standby "anyone free for a pass?"
+
+# Read — scope to one channel; optional --for filters out own sends.
+twapp msg fetch --channel reviewers
+twapp msg fetch --channel reviewers --for <me>
+
+# Observability — what channels exist, who's listening where.
+twapp msg channel list
+twapp msg channel subscribers reviewers
+```
+
+Subscription is by-convention: each worker declares what it listens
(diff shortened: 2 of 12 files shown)


piekstra (commented):

## Policy scan
Clean. `reviewer-standby` appears in docs/skills specifically to say "this idiom is deprecated — use channels instead", which is the right disposition.

## Correctness

### CLI surface — three interlocking changes

**1. `to: Option<String>`** — clap-level change. Was required; now optional so `twapp msg send --channel X "body"` parses without a fake positional. Runtime `cmd_send` rejects when neither `to` nor `--channel` produces a recipient. Tests:
- `clap_send_without_positional_to_parses_but_runtime_rejects` — clap accepts `send`, parsed fields are all `None`.
- `clap_send_accepts_channel_flag_with_single_positional` — `send --channel X "hello"` → `to=Some("hello")`, `channel=Some("X")`, `body=None` at the clap layer. The reinterpretation happens in `cmd_send`.

**2. Runtime reinterpretation of the single positional** — `channel_nonempty && body.is_none() → (None, to)`. That is: when `--channel X` is set and exactly one positional is present and no explicit body, treat the positional as the body. `channel_flag_alone_reroutes_single_positional_to_body` pins this E2E through `cmd_send` with `body: None` mimicking the clap-parsed shape. Subtle but correct.

The decision boundary leaves two less-obvious shapes untested:
- `send --channel X worker-a "body"` (two positionals + flag): clap fills `to=Some("worker-a")`, `body=Some("body")`. `body.is_none()` is false → no reinterpretation → multi-recipient send to `worker-a` AND `channel:X`. That's the intended behavior per the design's multi-recipient example ("direct + channel in the same send").
- `send "body"` with no `--channel` and no real recipient: `to=Some("body")`, body becomes stdin. Runtime sends to recipient named "body". Unchanged from pre-PR behavior.

Worth a one-liner comment or a third test pinning the "two positionals + channel flag" shape — would lock the intended "positional is recipient, not body" disambiguation against a future refactor.

**3. Filter reshape in `select_messages`** — the subtle one:
- When `channel` is set: apply `matches_channel` first, then `--for` (if set) downgrades to "exclude messages I sent" (`m.fm.from != h`).
- When `channel` is not set: preserve the original `matches_for_handle` semantics.

That's the right semantic split for the subscriber-fetch use case ("scope to channel X, but don't re-surface my own sends"). Without it, `matches_for_handle` would drop every channel message because `channel:X` isn't in the recipient slots it checks. `channel_fetch_with_for_excludes_self_sent_messages` pins this exact scenario with a two-sender fixture (`coordinator`, `reviewer-a`) — good coverage.

**4. Urgent-lane fast-path disabled when `--channel` is set** —
```rust
let use_urgent_fast_path = matches!(priority, Some(MsgPriority::Urgent) | Some(MsgPriority::Blocker)) && channel.is_none();
```
Correct defense: channel recipients deliberately don't get `urgent/` symlinks (`create_urgent_symlinks` skips `channel:*` recipients). Scanning the urgent lane for a channel fetch would return empty, silently hiding urgent channel messages. `fetch_priority_urgent_with_channel_filter_sees_channel_messages` lands this exact shape — urgent channel message + urgent direct + routine channel → returns only the urgent channel message. Good.

### Observability — `msg_channel.rs`

**`list_channels`** — reads `inbox/channel/*` subdirs, counts `.md` files per dir. `list_channels_includes_empty_channel_dir` pins the "empty dir still listed" behavior (useful for pre-declared channels). `list_channels_empty_inbox_returns_empty` for the cold path. Missing `inbox/channel/` → empty `Ok` vec via the `let Ok(entries) = ... else { return Vec::new() }` guard. Good.

**`channel_subscribers`** — filters `list_presence` by `claims.contains("channel:<name>")`. Relies on `list_presence` returning sorted-by-handle (the test `channel_subscribers_filters_by_claim` asserts `["alpha", "beta"]` for reviewers). **Potential concern**: if `list_presence`'s ordering ever changes, this output silently re-orders. A two-line `out.sort()` at the end of `channel_subscribers` would be defensive; not a blocker.

Dead handles (no presence file) never appear — correct.

### Test isolation consolidation (`test_env.rs`)

Clean refactor: five modules used to define their own `env_lock()` as a `OnceLock<Mutex<()>>` local to their `tests` module. Each lock was independent, so tests in `msg_claim` could race tests in `msg_presence` even though both mutate process-wide `TWAPP_MAILBOX_DIR`. `test_env::lock()` is now the single shared mutex every env-touching test acquires.

The `#![cfg(test)]` on the module gate keeps it out of non-test builds — no runtime cost. Poisoned-lock recovery via `unwrap_or_else(|p| p.into_inner())` ensures one panicking test can't brick the whole suite. 263/263 pass in parallel per PR body (was serial-only).

### Channel archive rotation pin
`channel_message_written_to_flat_archive_rotates_by_day_pr7` documents that channel messages rotate through PR-7's `archive rotate` exactly like broadcasts and directs do — no special handling needed. Good contract pin.

## Scope discipline
Files match the PR body:
- New: `msg_channel.rs` (+398), `test_env.rs` (+21)
- CLI wiring: `cli/mod.rs` (+9), `cli/msg.rs` (+356/-38 — mostly the `to: Option` and filter reshape)
- Test-env migration: `cli/coordinator.rs` / `msg_claim.rs` / `msg_cursors.rs` / `msg_migrate.rs` / `msg_presence.rs` (each -9 old lock, +3 uses)
- Docs: `README.md` (+56/-7), two skills updated (+41/-1, +15/-6)

No drift into PR-5 (presence) or the coordination PRs; scope matches design §2.3 PR-6 exactly. The test-env cleanup is a defensible side-quest — it had to happen to add a 6th `env_lock` without tipping the parallel-run flake, and consolidating is the right fix.

## Tests
- Rust: 12 new in channel/cmd_send axis (`channel_fetch_returns_only_messages_on_that_channel`, `channel_fetch_with_for_excludes_self_sent_messages`, `channel_fetch_on_unknown_channel_returns_empty_no_crash`, `channel_flag_alone_reroutes_single_positional_to_body`, `fetch_priority_urgent_with_channel_filter_sees_channel_messages`, `channel_message_written_to_flat_archive_rotates_by_day_pr7`, `list_channels_*`×3, `channel_subscribers_*`×2, `cmd_list_and_cmd_subscribers_exit_zero`) + the two clap-reshape tests. 263/263 pass.

Briefing's "Unknown channel fetch is a no-op" requirement is specifically locked by `channel_fetch_on_unknown_channel_returns_empty_no_crash` with a double-fixture (populated mailbox with other traffic AND a truly-empty mailbox). Good.

Gaps (not blockers):
- No test pinning "two positionals + `--channel`" = direct + channel multi-recipient. The CLI surface comment mentions this shape but it's not covered.
- No test for `channel_subscribers` sort stability under a hypothetical reorder of `list_presence`.
- No test for the `send --channel "" "body"` edge case (trimmed-empty channel flag). `channel_nonempty` check catches it, but explicit coverage would pin the "treat whitespace channel as absent" behavior.

## Docs consistency
- README "Channels (PR-6)" subsection is the right operator-facing description: send both forms, multi-destination direct+channel, broadcast-into-channel, fetch with `--for` semantics spelled out, observability commands, unknown-channel-fetch callout, archive-rotation note.
- Design coverage list in the main messaging paragraph now includes PR-6. Good.
- `skills/agent-coordinator/SKILL.md` Channels subsection is dense but high-signal. The deprecation callout (`> **Prefer channels over the -standby handle convention.**`) is load-bearing for the coordinator workflow — that idiom had been appearing in the archive for weeks. The archetype table's reviewer row updates from `reviewer-standby` to `reviewer (subscribes to channel:reviewers)` — right change.
- `skills/spawn-agent/SKILL.md` Example B gets the channel subscription pattern (`presence heartbeat --claims channel:reviewers` as step 1 of the loop). The example's clarity about "step 2 fetch the channel, step 3 process" is the right operator shape.

- The `cmd_send` reinterpretation block deserves a doc-comment at the function level summarizing the three-case decision table ("two positionals → recipient + body", "one positional + no channel → recipient, body from stdin", "one positional + channel → body, channel is the sole recipient"). Future maintainers will thank you.
- `channel_subscribers` could `.sorted()` defensively — one line, eliminates the silent-reorder risk if `list_presence` internals change.
- The deprecation-of-`-standby` call in the skill doc is strong but the reviewer-lane briefing I'm running from *right now* still uses the old `reviewer-standby` idiom in its text. If you want the deprecation to stick, a grep-and-update pass over `briefings/` (in the agent-collab repo, not twapp) would be a useful follow-up. Out of scope for this PR.
- `list_channels` counts symlinks as messages (comment is explicit about intent). If PR-3's split ever drops symlink shims into `inbox/channel/<name>/` the counts could inflate. Low-prob given current architecture.

Good, tightly-scoped PR. The three interlocking CLI changes (`to: Option`, runtime reinterpretation, filter reshape + urgent fast-path disable) are individually subtle but the test suite pins each seam. The test-env consolidation is a genuine parallel-hygiene win that was overdue. Ready to merge once CI stays green.

PR #66, opened 2026-04-21
fix(ui): hide urgent panel when no mailbox + inherit session theme

piekstra:
## Summary

Two user-reported bugs on the urgent-inbox sidebar panel from PR #57, folded into one diff since both touch the same component:

1. **Mailbox-gate bug.** Vanilla twapp instances (no `TWAPP_MAILBOX_DIR` / `TWAPP_SHARED_DIR`) were rendering a scary "URGENT: Error: No mailbox..." banner because the panel called `fetch_messages` unconditionally and displayed the backend error as a red styled row. Single-session users read the banner as a real urgent message.
2. **Theming bug.** The same panel hard-coded `background: var(--bg-primary)`, which renders white in light mode and ignores the per-instance session accent color the rest of the sidebar inherits. (Coordinator added this to scope 20260421T143500Z.)

### Fixes

- **New `get_mailbox_status` Tauri command** — pure filesystem probe that checks `TWAPP_MAILBOX_DIR` → `TWAPP_SHARED_DIR/mailbox/` → `<cwd>/mailbox/inbox/` and reports `configured: bool`. `UrgentInbox` calls it once at init and renders nothing until the probe resolves `configured: true`. No handle or no mailbox = no panel, no empty state, no error.
- **Discreet error footer** — `fetch_messages` failures now `console.warn` + show a small "Urgent feed unavailable" footer instead of the red banner. The `.urgent-error` style is left in the CSS as dead code; the component no longer reaches it.
- **Theme inheritance** — `.urgent-panel`, `.fleet-panel`, and `.timeline-panel` switch from `background: var(--bg-primary)` to `background: transparent`, so the sidebar's `--bg-secondary` (set by `applyThemeColor` to the session accent) shows through. Matches how notes / tickets already render.

### Audit of other wave-2/3 co-lab UI (per briefing)

- #52 session groups — `SessionLauncher.hasColabSections` already hides the "Co-lab groups" header when no session has `colab_group`. No leak.
- #63 fleet pane — `FleetPane` already returns `null` unless role === `coordinator`. No leak.
- #59 chrome — already gates on role / provenance. No leak.
- #55 launcher button — always-available per briefing, unchanged.
- #62 quick-action menu — `AgentContextMenu` ships in-tree but has no caller yet (only its own test imports it). Mailbox-gating deferred until a call site exists; flagged here as a **follow-up**.
- **Theming follow-up** — none of the audited components hard-code non-theme colors *except* the three panels fixed here. `.priority-chip.priority-blocker` and `.urgent-row-blocker` keep their literal red because red is semantic (blocker = red regardless of theme), same pattern `.config-error` uses.

## Test plan

- [x] `cargo test --lib msg::` — 91 pass, 8 new `probe_mailbox_status` cases covering every branch of the resolver (unset, empty-string env, MAILBOX_DIR hit, missing-dir fallthrough, SHARED_DIR fallback, SHARED_DIR without `mailbox/` subdir, cwd fallback, precedence).
- [x] `npm test` — 164 pass, 6 new `panelVisibility` cases. The three the briefing calls out: no-mailbox → hidden, configured + empty → empty-state, configured + message → list. Plus probe-pending → hidden, configured + error → discreet footer, list-wins-over-stale-error.
- [ ] Manual: unset `TWAPP_MAILBOX_DIR` and `TWAPP_SHARED_DIR`, launch an instance with a named handle → no urgent panel visible in the sidebar. Set `TWAPP_MAILBOX_DIR` to an existing mailbox with one urgent message → panel renders with the message. Same mailbox with zero urgent messages → "No urgent messages." empty state. (Requires `cargo tauri dev` which I can't run headless from this worktree.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -744,12 +744,18 @@ trace, not a crash.
 > top of any long write cycle — before the first line of code.
 
 Inside the session window, the sidebar shows an **Urgent** panel directly
-above Notes whenever the session has a handle. It polls `twapp msg fetch
---for <self> --priority urgent|blocker` every 10s, renders each message
-as a row (from + subject + priority chip + relative time), and auto-
-collapses after the queue has been empty for a minute. Click a row to
-open a read-only message view. Blockers get the strongest red accent,
-urgents a muted one. Single-session users with no handle see no panel.
+above Notes whenever the session has a handle **and a mailbox is
+configured** (i.e. `TWAPP_MAILBOX_DIR`, `TWAPP_SHARED_DIR/mailbox/`, or
+a `./mailbox/inbox/` directory under the session cwd resolves to an
+existing directory). It polls `twapp msg fetch --for <self> --priority
+urgent|blocker` every 10s, renders each message as a row (from + subject
++ priority chip + relative time), and auto-collapses after the queue has
+been empty for a minute. Click a row to open a read-only message view.
+Blockers get the strongest red accent, urgents a muted one.
+Single-session users with no handle — and regular twapp instances with
+no mailbox in sight — see no panel at all; a transient fetch failure
+shows a discreet "Urgent feed unavailable" footer instead of a red
+banner.
 
 #### Per-agent quick actions (context menu)
 
--- a/src-tauri/src/gui/mod.rs
+++ b/src-tauri/src/gui/mod.rs
@@ -130,6 +130,7 @@ pub fn run(args: GuiArgs) {
             files::reveal_in_finder,
             msg::send_message,
             msg::fetch_messages,
+            msg::get_mailbox_status,
             fleet::list_fleet,
             timeline::list_timeline_events,
             agent_actions::focus_agent_window,
--- a/src-tauri/src/gui/msg.rs
+++ b/src-tauri/src/gui/msg.rs
@@ -273,6 +273,14 @@ pub struct MailboxStatus {
 /// The briefing says: TWAPP_MAILBOX_DIR → TWAPP_SHARED_DIR/mailbox → `<cwd>/mailbox/inbox`.
 /// A path counts as "configured" only if it actually exists on disk — empty
 /// env strings, missing dirs, and non-dir files all fall through.
+///
+/// This is a *cheap* existence check, not an authoritative reachability one:
+/// a read-only inbox, a permission-stripped directory, or a stale NFS mount
+/// still pass the probe and will surface their real failure when
+/// `fetch_messages` runs. That's deliberate — the probe is there to spare
+/// vanilla single-session users any urgent-lane UI at all, not to guarantee
+/// the feed works. Fetch-time errors degrade to the "Urgent feed unavailable"
+/// footer rather than bouncing back up to a render-gate.
 pub fn probe_mailbox_status(
     mailbox_env: Option<&str>,
     shared_env: Option<&str>,
--- a/src-tauri/src/gui/msg.rs
+++ b/src-tauri/src/gui/msg.rs
@@ -6,11 +6,15 @@
 //! - `fetch_messages` — `msg fetch --format json`. Used by the urgent-inbox
 //!   panel (`src/components/UrgentInbox.tsx`), which calls it once per
 //!   priority (urgent + blocker) and merges the results.
+//! - `get_mailbox_status` — pure filesystem probe. Lets co-lab UI decide
+//!   whether to render at all. Vanilla single-session users never see the
+//!   urgent panel because `configured: false` short-circuits the render.
 
 use super::types::*;
 use serde::{Deserialize, Serialize};
 use serde_json::Value;
 use std::io::Write;
+use std::path::{Path, PathBuf};
 use std::process::{Command, Stdio};
 
 #[derive(Debug, Clone, Deserialize)]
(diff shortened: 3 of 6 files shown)


piekstra (commented):

## Policy scan
Clean. No private-project terms. The diff is UI-gating + theming, entirely in well-established vocabulary.

## Correctness

### Bug 1 — scary URGENT banner on vanilla sessions

**Root cause**: `UrgentInbox` called `fetch_messages` unconditionally when `selfHandle` was set. On vanilla single-session instances (no `TWAPP_MAILBOX_DIR`, no `TWAPP_SHARED_DIR`), the backend errored, and the error was rendered as a red `.urgent-error` banner. Users read it as a real urgent message. Confirmed cause.

**Fix shape** — mailbox probe at mount + panel gate:
- **`probe_mailbox_status`** is a pure function taking `(mailbox_env, shared_env, cwd)` — testable without env shim. Resolution order: `TWAPP_MAILBOX_DIR` → `TWAPP_SHARED_DIR/mailbox/` → `<cwd>/mailbox/inbox/`. Each leg requires `is_dir()` → so empty env strings, missing paths, and files-masquerading-as-dirs all fall through. Eight tests exercise every branch plus the precedence order (`mailbox_status_prefers_mailbox_env_over_shared_and_cwd` locks that `TWAPP_MAILBOX_DIR` wins when all three are configured).
- **`get_mailbox_status` Tauri command** reads env vars + `config.cwd` once per call, hands them to `probe_mailbox_status`. One-shot probe on mount (`useEffect` with `cancelled` flag), no polling. Right — env doesn't change mid-session in practice.
- **Panel gate**: `if (!mailboxConfigured) return null;` after the `selfHandle` check. `mailboxConfigured: boolean | null` with `null` meaning probe-in-flight → also returns `null` from the component, so no flash-then-hide. Good.

**`panelVisibility` pure helper** — separates gating math from the component so tests can pin all four outcomes without React. Seven cases covered including the two subtle ones:
- `"hides while probe is pending"` — `mailboxConfigured: null` suppresses render before resolution.
- `"prefers showing messages over the error footer when both are set (stale error)"` — if a later fetch succeeded after an earlier one failed, real messages win over the stale error. Prevents ghost-footer.

This is the kind of pure-function extraction that makes the regression test trivial — `panelVisibility("handle", false, msgs, null)` directly asserts "unconfigured hides even with populated messages," which is exactly the regression path.

### Bug 2 — panels hard-coded `--bg-primary`

**Root cause**: `.urgent-panel`, `.fleet-panel`, `.timeline-panel` used `background: var(--bg-primary)` which renders white in light mode and ignores `applyThemeColor`'s `--bg-secondary` session accent.

**Fix**: switch to `background: transparent` on all three. Tones (e.g. `.urgent-tone-blocker`) keep their red overlay via `rgba()` which composites correctly over whatever sidebar accent is behind. The `.priority-chip.priority-blocker` / `.urgent-row-blocker` literals stay red — correctly preserved because red is *semantic* (blocker = red regardless of theme), not chrome.

The PR body's comment "same pattern .config-error uses" is the right framing — themeable chrome vs. semantic color, the line is drawn correctly.

### Discreet error footer
- `.urgent-error` CSS class is still defined but no longer used (dead CSS per PR body). Not a blocker but worth a follow-up cleanup to delete the dead rule.
- New `.urgent-footer-unavailable` is italic + muted + small + centered — anti-banner visual language. `title={error}` surfaces the detail on hover for operators who want it. Good trade-off: noisy enough to signal "something's not right" without triggering the "this is an urgent message!" misread.
- `console.warn("[urgent] fetch failed:", e)` before `setError` means operators with devtools open still see the underlying cause. Right balance.

### Audit of other co-lab panels

The PR body's audit matches what I can verify:
- `FleetPane` — gates on `isCoordinator`, returns `null` otherwise. No mailbox leak.
- `TimelinePane` — same `isCoordinator` gate. No leak.
- `SessionLauncher` session-groups — `hasColabSections` already hides the Co-lab header when empty.
- `AgentContextMenu` (from PR #62) — no caller in-tree yet (only its own test imports it). The audit correctly defers mailbox-gating to whenever a call site wires up, and flags it in the PR body. That's the right disposition.
- Launcher button (#55) — always-available per briefing; nothing to gate.

## Scope discipline
Files match the PR body: `gui/{mod,msg}.rs` (+1, +153), `App.css` (+21/-5), `UrgentInbox.{tsx,test.ts}` (+80/-7, +66), README (+12/-6). 333/18 total. Tightly scoped.

The three-line CSS change on fleet + timeline panels is perfectly in-scope — both use `UrgentInbox`'s exact pattern of `bg-primary → transparent`. Audit noted the same theming bug on all three panels; fixing them together is right.

## Tests
- Rust: 8 new `probe_mailbox_status` cases covering every resolver branch + precedence + empty env strings + missing-dir fallthrough. 91/91 pass per PR body.
- TS: 6 new `panelVisibility` cases covering all four `PanelVisibility` variants + the stale-error-vs-list precedence. 164/164 pass per PR body.

Gaps (not blockers):
- No test exercises the actual `get_mailbox_status` Tauri command (env reads, `config.cwd` wiring). The pure `probe_mailbox_status` is well-covered; the Tauri wrapper is trivial enough that this is fine, but one `#[test]` setting env vars + asserting against an injected `GuiArgs` would close the loop.
- No test for the `cancelled` flag in the mount-probe useEffect. A vitest with a delayed fetcher resolve + immediate unmount would catch "setState on unmounted component" regressions.
- No test for the probe-failure path (`.catch(() => setMailboxConfigured(false))`). The soft-failure behavior is the right disposition but not test-pinned.

## Docs consistency
README section on urgent panel rewritten to call out the mailbox-configured gate: `TWAPP_MAILBOX_DIR → TWAPP_SHARED_DIR/mailbox/ → ./mailbox/inbox/`. Users reading "why isn't the panel showing?" will find their answer. Good.

- `.urgent-error` CSS class is now dead. Deleting it + its comment prevents a future contributor from re-introducing the banner thinking the style is still in use.
- `probe_mailbox_status`'s `cwd` argument is `Option<&Path>` while the env args are `Option<&str>`. That's fine today (Tauri command layer has its own conversion), but if the contract ever needs to be called from more places, a consistent `AsRef<Path>` bound would simplify the call sites.
- The `configured: true` check only validates directory existence, not writability / readability. A read-only mailbox or a permission-stripped inbox would still pass the probe and trigger the discreet footer on first fetch. That's correct graceful-degrade behavior for today, but worth a comment noting the probe is *cheap*, not *authoritative*.
- `mailbox_env.map(str::trim).filter(|s| !s.is_empty())` is elegant; the `is_dir()` check that follows doesn't try to handle symlinks-to-dirs explicitly (which `is_dir()` resolves). Documented behavior — `std::fs::metadata` follows symlinks by default. Fine.

Tight, well-tested bug-fix + theming PR. The pure-function extractions (`probe_mailbox_status`, `panelVisibility`) are exactly the right shape for regression tests; the audit of other panels closes the "does this bug hide elsewhere?" question; the theming fix touches the three panels that needed it and nothing more. CI is green, merge state is CLEAN. Ready to ship.

PR #71, opened 2026-04-25
fix(ui): react to live role changes + gate Send Message

piekstra:
## Summary
- `App.tsx` now re-reads `get_session_info` on a 2s poll so role changes
  flowing from `twapp coordinator claim` / `launch` / external edits
  propagate to the live window without a reload. Fleet pane reacts to
  role flips in ~2s. `setSessionRole` / `setSessionColabGroup` use
  functional updates that skip the set when the value is unchanged so
  React doesn't re-render spuriously.
- Send Message button hidden on sessions whose role is null / "" / "-"
  (sessions that don't participate in the mailbox). Extracted as a pure
  `canSendMessage(role)` helper in `src/utils/colab.ts` with vitest
  coverage covering null / undefined / empty / "-" / coordinator /
  implementer / reviewer / log-watcher (forward-compat).

## Notes / scope
- Replaced (rather than augmented) the original one-shot `get_session_info`
  call inside the giant mount effect — the new polling effect does the
  initial load itself, so no double-load on mount.
- Did NOT gate the Cmd+Shift+M keyboard shortcut at App.tsx:986-990
  (which is the second entrance to the composer). Doing so requires
  adding `sessionRole` to that handler's dependency array (currently
  `[tabs, activeTabId]`), which felt like spread beyond the briefing's
  scope. Easy follow-up if desired.
- Backend (`src-tauri/`) untouched — `get_session_info` already returns
  the raw session JSON.

## Test plan
- [x] `npx tsc --noEmit` clean
- [x] `npm test` passes — 167 tests, including 12 in `colab.test.ts`
      (3 new `it` blocks covering `canSendMessage`)
- [x] `npm run dev` serves on http://localhost:1420; Vite transforms
      App.tsx without errors (480KB output, HTTP 200). Cannot drive
      the running window from here, so the manual "flip
      `.twapp-session.json` and watch Fleet appear" verification is
      left for the reviewer / next dev session.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -23,6 +23,7 @@ import { isYamlFile, isHtmlFile, isImageFile, imageMimeType, isFilePath, isLikel
 import { remarkAutolinkFilePaths } from "./utils/markdown";
 import { buildResumeCommand } from "./utils/session";
 import { isNewerVersion } from "./utils/version";
+import { canSendMessage } from "./utils/colab";
 import { renderJsonNode, renderYamlNode } from "./components/FilePreview/renderers";
 import PromptSections from "./components/PromptSections";
 import type { EditingPromptState } from "./components/PromptSections";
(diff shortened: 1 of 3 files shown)


piekstra (commented):
Ship it.

- Polling effect: `cancelled` flag guards post-await setState, `clearInterval` in cleanup, empty deps, functional setters avoid spurious re-renders. No leak.
- `canSendMessage`: pure, correctly rejects null/undefined/""/"-", forward-compat for new role values, well-tested.
- Send Message button gate is the only affordance touched; Cmd+Shift+M shortcut deliberately left as a follow-up.
- Scope tight: only `App.tsx`, `colab.ts`, `colab.test.ts`. No backend or unrelated component churn.
- 2s poll interval is fine — file is tiny, role changes rare.

